PR #26, opened 2026-03-04
update runtimelib and jupyter-protocol

rgbkrk:
This upgrades `runtimelib` from 1.0 to 1.4 and upgrades `jupyter-protocol` from 1.0 to 1.4.

You can view the [full `jupyter-protocol` crate CHANGELOG on GitHub](https://github.com/runtimed/runtimed/blob/main/crates/jupyter-protocol/CHANGELOG.md). The main changes are improved compatibility with kernels that omit or deivate from spec required fields. There's more defensive deserialization for those cases.

On the runtimelib side, `.split()` is now supported on the Jupyter shell channel to make it so reader and writer can be owned separately. I haven't looked at your code deeply enough to see if you can make use of it. Within Zed and Deno it made the code cleaner and allowed us to drop extra tasks and channels.

On the nbformat side, it now supports v3 notebook parsing and upconversion to v4. There are also some fixes to handle both string and array for cell source (noticed with how JetBrains writes out notebooks, still valid nbformat).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4116,9 +4116,9 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 
 [[package]]
 name = "sdd"
-version = "4.7.1"
+version = "4.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


cscheid:
Thanks!

PR #46, opened 2026-03-16
Replay for Quarto Hub

shikokuchuo:
This PR adds **replay** to Quarto Hub collaborative editor.

This allows you to explore and step through the full document history, and provides an interface to automerge-native versioning / undo.

Adds a Replay button along the bottom edge of the editor. Clicking it opens a drawer and enters Replay mode. Replay mode is a personal exploration space - it doesn't affect other collaborators. Syncs continue in the background, and exiting replay returns you to the live collaborative document.

The drawer contains playback controls to play through the document's history like a video, or scrub back and forth along the timeline. The timeline scrubber shows a visual "waveform" - coloured bars representing different authors, so you can see at a glance when different collaborators contributed.

At any point during replay, hit the Restore button to revert the document to that moment in time - this adds a new change to the document. To undo this, simply go back into replay mode, step back one frame, and click Restore again.

There are keyboard shortcuts (spacebar, arrow keys, Home/End, Escape) to navigate without clicking.

[video.webm](https://github.com/user-attachments/assets/f853cf46-6d60-420d-be78-943952b8c3c8)

### DIFF
--- a/hub-client/src/components/ReplayDrawer.test.tsx
+++ b/hub-client/src/components/ReplayDrawer.test.tsx
@@ -19,7 +19,7 @@ function makeState(overrides: Partial<ReplayState> = {}): ReplayState {
     currentContent: '',
     timestamp: null,
     actor: null,
-    chunkAuthors: [],
+    chunkActors: [],
     ...overrides,
   };
 }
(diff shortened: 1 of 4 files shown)


cscheid:
This is _excellent_.

It would be really useful to have corresponding replay APIs for `quarto-sync-client`, I think. We should merge this right away for us to play with it and showcase at the i&s workshop, but I wonder if we could refactor `hub-client/src/hooks/useReplayMode.ts` into something that's usable directly via `quarto-sync-client`.

Another useful thing would be a way for this mode to work on changes to the index document. There, the UI gets a bit more complicated. Users never see the "content" of the index automerge document and don't know the representation, and so it's not trivial to show the equivalent of the editor and preview. The main thing we want to allow is for people to undelete files, but there are more complicated scenarios, like the fact that someone might have deleted a file from the project, then recreated a new file with the same name (but different automerge id)

PR #80, opened 2026-03-23
Fix: prevent remote edits from being silently deleted on concurrent updates

### LINKED ISSUE #74: hub: edits can get lost
You can get into a bad state where you type and then it disappears. On a collaborative session for Shiny team meetings (7 users).

Possible history with no changes? To confirm.

Switching and coming back fixes the bad state.

shikokuchuo:
Fixes #74.

When two people edit the same file at the same time, the second person's changes could get silently thrown away.                   

What was happening: When you type in the editor, two things happen: (1) your edit gets merged into the shared document via Automerge, and (2) the app updates its local copy of the file content. The problem was that step 2 was overwriting the local copy with just your text — ignoring any changes the other person made in the meantime. The next time you typed, the app would compare against that stale copy and effectively erase the other person's work.

We removed the redundant local state update in App.tsx and the redundant callback in the sync client. Both were unnecessary because the Automerge change handler already pushes the correctly-merged document state back to the app. Now there's only one source of truth for file content — the merged Automerge document — so remote edits can't get lost.

### DIFF
--- a/hub-client/src/App.tsx
+++ b/hub-client/src/App.tsx
@@ -373,12 +373,14 @@ function App() {
   }, [navigateToProjectSelector]);
 
   const handleContentChange = useCallback((path: string, content: string) => {
+    // updateFileContent fires handle.change(), which synchronously triggers the
+    // 'change' event on the DocHandle. The registered changeHandler calls
+    // callbacks.onFileChanged with the true merged Automerge document state,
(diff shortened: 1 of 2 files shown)


vezwork:
I looked at the change and your explanation and it makes a lot of sense to me fwiw. Thanks for adding comments that explain the flow.

PR #94, opened 2026-03-31
OT-based cursor tracking for collaborative presence

### LINKED ISSUE #9: Bug: Cursor positions of other users as I type feels like it should be anchored for each line
When I type, the other team mates cursors appear. Np!

However, as I type, their cursor positions are "x many characters ahead" of my newly added text, possibly wrapping to the next line of text.

While this a fair place to start, I feel that it is only valid when the cursors are within the same line. 

---------

Goal: When I type and other cursors are on different lines, I expect their cursor to be consistently located within their line and to not leave that line. Almost as if any line breaks will stop any horizontal movement of their cursor.

--------------------

Add lines:

https://github.com/user-attachments/assets/977f22e8-faee-471b-9789-0cdef0c7a7ce

Delete lines:

https://github.com/user-attachments/assets/486b8d36-0473-4b35-ad72-cb0454f2c695

shikokuchuo:
Fixes #9. Deletion in one paragraph no longer shifts remote cursors in subsequent paragraphs.

Remote cursor positions are now adjusted locally via operational transformation (shifting stored character offsets by the length delta of each edit, based on whether the edit falls before, inside, or after the cursor) instead of being rendered from stale network offsets. When the document changes (local or remote edits), each tracked cursor/selection offset is shifted accordingly, keeping decorations at their correct logical position between presence updates.

Key details:
- `PeerCursorState` tracks OT-adjusted offsets and last-raw-presence values per peer, so we can distinguish a genuinely new presence message from a re-render of an already-shifted value
- `anticipatingEditRef` prevents double-shifting when a presence message arrives before its corresponding document change (e.g. a remote user types and we receive their new cursor position before the inserted character syncs): restricted to small forward cursor movements to avoid false positives from navigation clicks
- Removes the old pre-compensation heuristic (`prevCursorsRef`, `cursorModelLengthRef`, `MAX_TYPING_DELTA`)

### DIFF
--- a/hub-client/src/hooks/usePresence.ts
+++ b/hub-client/src/hooks/usePresence.ts
@@ -78,6 +78,23 @@ function ensureCursorStyle(color: string, odorId: string): void {
 }
 
 /**
+ * Shift a character offset to account for a single edit.
+ * Offsets after the edit move by delta; offsets inside the replaced range
+ * clamp to the end of the replacement text; offsets before are unchanged.
+ */
+function transformOffset(
+  offset: number,
+  editStart: number,
+  oldEnd: number,
+  newEnd: number,
+  delta: number,
+): number {
+  if (offset >= oldEnd) return offset + delta;
+  if (offset > editStart) return newEnd;
+  return offset;
+}
+
+/**
  * Convert a color to a safe CSS class identifier.
  */
 function colorToId(color: string): string {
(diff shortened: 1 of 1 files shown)


vezwork:
Seems like a good solution. Nit: Instead, I think it may reduce code and be a bit more robust to use https://automerge.org/automerge/automerge/struct.Cursor.html

edit: made an issue for this nitpick: https://github.com/quarto-dev/q2/issues/113

PR #109, opened 2026-04-09
Add real wasm32 smoke tests and CI job

cderv:
Replace the lost cfg proxy coverage (removed in #116) with 6 real WASM smoke tests
in `crates/pampa/tests/wasm_lua.rs` that compile and run on `wasm32-unknown-unknown`:

- Restricted Lua VM creation (`Lua::new_with()`)
- Filter execution through the WASM code path (VFS + synthetic io/os)
- Shortcode engine initialization
- Error handling (validates `panic_unwind` works)
- Synthetic `io` and `os` module registration

These run in a new `wasm-tests` CI job (Linux, nightly + Clang).

## Infrastructure

- `crates/wasm-c-shim/` — shared C stdlib shim crate for wasm32 (no libc on wasm32,
  C deps like tree-sitter and Lua need `malloc`, `fprintf`, etc. at link time)
- `.cargo/config.toml` — wasm-bindgen-test runner config
- Gate `proptest`, `insta`, `tempfile`, `tokio` dev-deps on `not(wasm32)` to avoid
  pulling `getrandom 0.3.4` which doesn't compile for wasm32
- Gate pampa `[[bin]]` targets via `required-features` to prevent compilation during
  WASM test runs (rust-lang/cargo#12980)
- Use `RUSTUP_TOOLCHAIN=nightly` in CI to bypass `rust-toolchain.toml` sysroot conflict
  with `-Zbuild-std`

## Known blocker

The `wasm-tests` CI job currently fails because `quarto-system-runtime/src/wasm.rs`
has 4 `raw_module` extern blocks (`/src/wasm-js-bridge/{template,sass,cache,fetch}.js`)
that get baked into any wasm32 binary. `wasm-bindgen-test-runner` generates `require()`
calls for these absolute paths, which fail in Node.js.

Recommended fix: feature-gate the JS bridge in `quarto-system-runtime` — add a
`js-bridge` feature, gate the 4 extern blocks, provide stubs when off.
`wasm-quarto-hub-client` enables `js-bridge`; pampa test builds don't.

Depends on #116.


gordonwoodhull:
@cderv, thanks for writing these tests! 

I was trying to understand why these paths weren't tested, and either misunderstood or miscommunicated this to Claude, testing them the wrong way! This was the mistake you fixed in
* #116 

I've fixed the remaining issues with this PR and it runs clean here:
https://github.com/quarto-dev/q2/actions/runs/24612984662 

I also rebased onto latest main. May I force push, or do you have other work on this branch?

cderv:
Oh thanks for chiming in ! I was going to continue with this this week. It needed to be rebased on main and it is also related to work I started in #125 

My aim was that we have some wasm build test, but I was sorting out quarto-hub build from other wasm build. 

interested to see your change ! 

Yes you can force push here. And maybe we can discuss quickly all this together if you have time today.

gordonwoodhull:
Thanks! I've pushed. It's really just

* the feature-gate fix you suggested above
* build flags enabling the same error handling / panic strategy for `wasm32-unknown-unknown` tests as used for the lua-wasm build in `wasm-quarto-hub-client`
* merge/rebase conflict resolution
* docs

(I'm going for a hike today since I stupidly worked over the weekend, glad to discuss later in the week.)

gordonwoodhull:
I dropped the unrelated stuff from
* #125 
(wasm-qmd-parser and rust toolchain)

and squashed this. I'll go ahead and merge it now.

Thanks @cderv!

PR #116, opened 2026-04-14
Fix WASM cfg proxy: remove test arm from wasm32 guards

cderv:
The `#[cfg(any(target_arch = "wasm32", test))]` pattern in `filter.rs` and
`shortcode.rs` forced native tests through the WASM-restricted Lua stdlib.
This caused Windows test failures because the restricted stdlib (no io/os/debug)
doesn't work with Lua's C implementation on native targets.

## Fix

Remove the `test` arm from the cfg guards so native tests use `Lua::new()` with
the full stdlib. WASM coverage is provided by dedicated WASM tests in CI.

Also marks the `dofile_wasm::test_dofile_script_dir_stack` test as
`#[cfg_attr(not(target_arch = "wasm32"), ignore)]` since it depends on
`register_wasm_dofile` which is only available in the WASM build.

## Changed files

- `crates/pampa/src/lua/filter.rs` — cfg guard fix
- `crates/pampa/src/lua/shortcode.rs` — cfg guard fix
- `crates/pampa/src/lua/dofile_wasm.rs` — ignore test on native
- `.claude/rules/wasm.md` — document the prohibited cfg pattern
- `claude-notes/instructions/testing.md` — update testing docs
- `dev-docs/wasm.md` — update WASM architecture docs

### DIFF
--- a/.claude/rules/wasm.md
+++ b/.claude/rules/wasm.md
@@ -1,0 +1,15 @@
+# WASM Code Rules
+
+## Never add `test` to wasm32 cfg guards
+
+The cfg pattern `#[cfg(any(target_arch = "wasm32", test))]` is prohibited. It forces
+native tests through the WASM-restricted Lua stdlib, which fails on Windows.
+
+Correct pattern:
+```rust
+#[cfg(target_arch = "wasm32")]
+// WASM-specific code (restricted Lua stdlib, synthetic io/os)
+
+#[cfg(not(target_arch = "wasm32"))]
+// Native code (full Lua stdlib via Lua::new())
+```
--- a/.claude/rules/xtask.md
+++ b/.claude/rules/xtask.md
@@ -19,9 +19,15 @@ paths:
 
 | Command | Alias | Purpose |
 |---------|-------|---------|
-| `cargo xtask dev-setup` | `cargo dev-setup` | Install required dev tools (cargo-nextest, wasm-pack) |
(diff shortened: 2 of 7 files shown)


vezwork:
@cscheid I'm gonna want something like wasm-qmd-parser again in the future for the Quarto extension, but putting that off for now.

cderv:
Let's leave it then - I was just trying to cleanup while understanding things around and fixing the problem I found.

cderv:
Done — I've kept the crate skeleton (Cargo.toml, src/, tests/, licenses) and cleaned up the stale parts (old CI configs, C shim superseded by `wasm-c-shim`, wasm-sysroot headers, wasm-pack demo artifacts). The README now documents it as dormant.

Edit: I’ll probably split work on this. No need to cleanup if you plan to rework on it. But this is definitely stale.

cderv:
Split this PR — the wasm-qmd-parser cleanup and workflow modernization are now in #125. This PR is just the cfg guard fix (removing the `test` arm from `#[cfg(any(target_arch = "wasm32", test))]`).

gordonwoodhull:
Thanks @cderv! 

I've gone back and verified that adding `test` was a misunderstanding. The intent was to test `io` and `os` using the wasm runtime, not to make native tests use the wasm runtime!

gordonwoodhull (approved):

PR #122, opened 2026-04-15
Add per-node attribution to q2-debug AST view

### LINKED ISSUE #115: Replay UI: show attribution of highlighted text
Could we highlight a section of text and then HIGHLIGHT the last time it was changed and by who on the changelog bar across the bottom.

shikokuchuo:
Closes #115.

## Summary

Adds per-node authorship to the q2-debug AST view. Each node is coloured by the author who last edited the source bytes behind it, with a hover badge showing the author and a relative timestamp. Off by default — toggle "Authorship" in the Settings sidebar (q2-debug only).

## How Automerge attribution meets the AST

Every AST node carries a `sourceInfoId` (the `s` field) that points into the parse's `sourceInfoPool`. `SourceInfoReconstructor` (from `@quarto/annotated-qmd`) resolves that id to a byte range `{ fileId, start, end }` in the original source.

Separately, an **attribution producer** replays Automerge history and records, per character, who wrote it and when. Producers expose only a minimal query interface:

```ts
interface AttributionSource {
  queryByteRange(fileId, byteStart, byteEnd): { actor, time } | null;
}
```

Attributing a node is then: resolve its byte range via the reconstructor, hand that range to the source, and (for multi-byte ranges) take the most recent writer — "last touch wins". The interface is byte-indexed (UTF-8, as emitted by WASM); the Automerge-backed source keeps an internal byte→char map since Automerge text is UTF-16.

## Cold start vs incremental updates

`useAttribution(filePath, sourceText)` owns the build lifecycle:

- **Cold start**: `buildRunListAttribution` replays full Automerge history, 500 entries per chunk, yielding via `requestIdleCallback` before every chunk. The hook returns `null` while the build is pending, so the document paints un-attributed first.
- **Incremental**: on each `sourceText` change (debounced 500 ms), `updateRunListAttribution` applies only the new history entries to the existing runs synchronously. On `HistoryCompactedError`, we fall back to a full rebuild.
- Switching files aborts any in-flight build and clears state before the next cold start.

`useNodeAttributionResolver(astContext, attributionCtx)` then wraps the source + reconstructor into a memoised, cached `getNodeAttribution(sid) → { actor, time, color, name } | null`. `color` and `name` come from an `identities` map supplied by the Automerge sync layer; unmapped actors fall back to a short hex of the actor ID.

## Rendering: text colour and hover tooltip

A node with a resolved attribution is wrapped in `<span|div class="q2-attr-wrap" data-sid="…" style="color: <actor-color>">`. The colour is plain inline style — no registry magic.

Hover is **event-delegated**: a single `onMouseOver`/`onMouseOut` pair on the container finds the nearest `.q2-attr-wrap`, reads `data-sid`, calls `getNodeAttribution`, and positions one floating `<AttributionBadge>` via `getBoundingClientRect`. One handler, one badge — not N of each.

## Producer-swappable design (git-blame adapter)

`AttributionSource` is the only boundary that matters: anything above it (storage shape, producer) can change without touching anything below (hook, resolver, renderer).

The default producer is Automerge-backed and run-length encoded (`attribution-runs.ts`) — 4× faster updates, 5× faster queries, 20× smaller than the per-char prototype (`attribution.ts`) on realistic batched workloads.

`attribution-gitblame.ts` is a second, independent producer that parses `git blame --porcelain`, converts lines to byte-ranged runs using `TextEncoder` (so multi-byte UTF-8 lands on correct byte offsets), and returns an `AttributionSource` with binary-search queries. It touches **no consumer code** — concrete proof the boundary holds and Automerge is swappable.

## Access from render components

`Editor.tsx` publishes `AttributionContext` (`{ source, identities, sourceText }`) to the render tree whenever attribution is on. Any renderer can opt in:

1. `useContext(AttributionContext)` to get the source.
2. `useNodeAttributionResolver(ast.astContext, attributionCtx)` for a `getNodeAttribution(sid)` function.
3. Read `sid` from a node's `s` field and look up `{ actor, time, color, name }`.

`ReactAstDebugRenderer` is currently the only consumer; any other renderer picks this up in three lines.

## Enabling the feature

A persisted `attributionEnabled` preference (default `false`) is exposed as an "Authorship" toggle in the Settings sidebar, shown only when the current format is `q2-debug`.

`Editor.tsx` passes `filePath = null` to `useAttribution` when the toggle is off, which short-circuits the history replay — zero cost when disabled, and flipping off mid-session aborts any in-flight build.

## Other notes

- Minor type fixes in `ts-packages/annotated-qmd` so it compiles cleanly under hub-client's strict tsconfig.


shikokuchuo:
After discussion with Carlos we're going to move this to a transform step of the render pipeline, so attribution will be available to all consumers - cli as well as hub-client.

shikokuchuo:
Closing in favour of #190.

PR #129, opened 2026-04-21
Replace OT cursor tracking with Automerge cursors in presence

### LINKED ISSUE #113: nitpick refactor: Use automerge cursor instead of OT for multiplayer cursor tracking
https://github.com/quarto-dev/q2/pull/94 seems like a good solution for cursor tracking.... but..... Nit: Instead, I think it may simplify code and be a bit more robust to use https://automerge.org/automerge/automerge/struct.Cursor.html

This would mean creating a automerge cursor from position whenever we receive a multiplayer cursor update and letting it hang about, syncing monaco cursors to it, until another multiplayer cursor update comes through and we replace the automerge cursor with a new one.

shikokuchuo:
Fixes #113.

Remote cursor/selection positions now travel as Automerge cursor strings anchored to `['text']` and resolve on the receiver against its local doc. Deletes the hand-rolled OT machinery in `usePresence.ts` (`transformOffset`, `peerStateRef`, `anticipatingEditRef`, same-line guard, OT `onDidChangeContent` effect); `modelVersion` stays as a re-render trigger.

### DIFF
--- a/claude-notes/plans/2026-04-21-automerge-cursors-for-presence.md
+++ b/claude-notes/plans/2026-04-21-automerge-cursors-for-presence.md
@@ -1,0 +1,481 @@
+# Replace OT cursor tracking with Automerge cursors in hub-client presence
+
+Beads: TBD (tracks [issue #113](https://github.com/quarto-dev/q2/issues/113))
+
+## Overview
+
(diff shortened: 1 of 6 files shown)


cscheid:
Excellent!

PR #139, opened 2026-04-28
Fix CRLF pipe-table line ending in tree-sitter scanner

### LINKED ISSUE #138: Pipe tables with CRLF line endings absorb the following block
A qmd document with CRLF line endings, a pipe table, and a paragraph immediately after the table parses incorrectly: the table absorbs the following paragraph as a malformed body row instead of treating it as a separate block. The same input with LF parses correctly.

This affects any Windows user with default `core.autocrlf=true` (the recommended Windows install default for git). The bug is silent — no error surfaces, just structurally wrong output in every pampa writer (HTML, PDF, Word, etc.).

I reproduced this on Windows at 80f07171.

## Reproducer

Two layers. Layer 1 hits `grammar.js` directly; Layer 2 hits `cargo run -p pampa` end-to-end and confirms the bug propagates to the Pandoc AST that drives every output writer (pampa does not normalize line endings before parsing).

```bash
# Fixtures
printf 'before\r\n\r\n| a | b |\r\n|---|---|\r\n| 1 | 2 |\r\n\r\nafter\r\n' > crates/tree-sitter-qmd/tree-sitter-markdown/test-crlf.qmd
printf 'before\n\n| a | b |\n|---|---|\n| 1 | 2 |\n\nafter\n' > crates/tree-sitter-qmd/tree-sitter-markdown/test-lf.qmd

# Layer 1 — grammar.js directly
cd crates/tree-sitter-qmd/tree-sitter-markdown
tree-sitter parse test-crlf.qmd > test-crlf.parse
tree-sitter parse test-lf.qmd > test-lf.parse
diff test-crlf.parse test-lf.parse

# Layer 2 — pampa end-to-end (default -t native)
cd ../../..
cargo run --quiet -p pampa --bin pampa -- crates/tree-sitter-qmd/tree-sitter-markdown/test-crlf.qmd > pampa-crlf.out
cargo run --quiet -p pampa --bin pampa -- crates/tree-sitter-qmd/tree-sitter-markdown/test-lf.qmd  > pampa-lf.out
diff pampa-crlf.out pampa-lf.out
```

## Observed parse trees (Layer 1)

LF — correct, table closes at the blank line:

```
pipe_table [2, 0] - [5, 0]
  pipe_table_header [2, 0] - [2, 9] ...
  pipe_table_delimiter_row [3, 0] - [3, 9] ...
  pipe_table_row [4, 0] - [4, 9] ...
pandoc_paragraph [6, 0] - [7, 0]
  pandoc_str [6, 0] - [6, 5]
```

CRLF — buggy, table extends through the blank line and absorbs the next paragraph:

```
pipe_table [2, 0] - [7, 0]
  pipe_table_header [2, 0] - [2, 9] ...
  pipe_table_delimiter_row [3, 0] - [3, 9] ...
  pipe_table_row [4, 0] - [4, 9] ...
  pipe_table_row [6, 0] - [6, 5]   <- spurious; should be a separate paragraph
    pipe_table_cell [6, 0] - [6, 5]
      pandoc_str [6, 0] - [6, 5]
```

## Observed Pandoc AST (Layer 2)

```
LF:   [Para "before", Table ... [body rows: [1, 2]],                Para "after"]
CRLF: [Para "before", Table ... [body rows: [1, 2], [Plain "after"]]]
```

CRLF case loses the trailing `Para`; the "after" paragraph becomes a one-cell table row.

## Likely location

In the external scanner's `PIPE_TABLE_LINE_ENDING` emission and the matching pipe-table newline rules in `grammar.js`.

https://github.com/quarto-dev/q2/blob/80f07171a902810e0a9c6a2029b05e0d0effc494/crates/tree-sitter-qmd/tree-sitter-markdown/src/scanner.c#L2256-L2261

https://github.com/quarto-dev/q2/blob/80f07171a902810e0a9c6a2029b05e0d0effc494/crates/tree-sitter-qmd/tree-sitter-markdown/grammar.js#L232-L245

Line 2259 in `scanner.c` reads `(lexer->lookahead == '\n' || lexer->lookahead != '\r')` — the second clause looks like a typo of `==`, since as written the OR is true for nearly any input. The inverse condition just above on line 2256 uses `!= '\n' && != '\r'`, which is consistent with `==` being intended on line 2259.

This grammar is forked from `tree-sitter-markdown`. We could check upstream for an existing fix before patching locally.

## How this surfaced

`cargo xtask verify` step 4 (`cd crates/tree-sitter-qmd/tree-sitter-markdown && tree-sitter test`) fails 5/451 tests on Windows: Example 201 (GFM), "Pipe table can precede content" (two variants), "Pipe table can follow content", "Pipe table with blank-line caption". Linux CI passes for the same commit because corpus files are LF on Linux checkout (`git ls-files --eol` confirms `i/lf w/crlf` for `test/corpus/pipe_table.txt`).

## Notes

- A `.gitattributes` entry pinning `test/corpus/*.txt` to LF would make the corpus tests pass on Windows, but it would only mask the symptom — the corpus failures are evidence of a real grammar bug, not corpus formatting.
- We could keep CRLF in the test corpus on Windows checkout (free coverage once the grammar is fixed), or pin the corpus to LF and add a separate Windows-only regression test.
- We could also audit `grammar.js` and `scanner.c` for other constructs that may assume LF only — pipe tables surfaced first because the corpus exercises them, but other block paths could be affected.

cderv:
When a Windows checkout (default `core.autocrlf=true`) authors a qmd document with a pipe table followed by content, the table absorbs the trailing paragraph as a malformed body row instead of closing at the blank line. Surfaces as 5 failing pipe-table tests under `tree-sitter test` on Windows; Linux CI passes even with the bug present.

## Root Cause

In `crates/tree-sitter-qmd/tree-sitter-markdown/src/scanner.c`, line 2259 had a typo. Paired with line 2256, the two checks are meant to be opposites:

- 2256: lookahead is NOT a newline (content follows) -> emit `PIPE_TABLE_LINE_ENDING` (table continues with another row).
- 2259: lookahead IS a newline (blank line follows) -> emit `LINE_ENDING` (table closes).

The expression `(lookahead == '\n' || lookahead != '\r')` is logically `lookahead != '\r'` and never matches `\r`. With LF the first clause matched `\n` and the table closed correctly by accident. With CRLF, the lookahead after consuming `\r\n` is `\r` (start of the blank line); the buggy expression rejected it, so fall-through emitted `PIPE_TABLE_LINE_ENDING` via line 2287 instead — extending the table past the blank line.

The buggy block was introduced in 79489c4e on this fork. All other newline checks in `scanner.c` already use the paired `'\n' || '\r'` form.

## Fix

Change `!=` to `==`. The 5 originally-failing pipe-table corpus tests now pass on Windows: Example 201 (GFM), Pipe table can precede content (two variants), Pipe table can follow content, Pipe table with blank-line caption.

## Tests

Two layers of regression coverage:

- A Rust unit test `pipe_table_crlf_matches_lf` that builds the CRLF input in-process and asserts the parsed S-expression matches the LF parse. The existing corpus tests miss this on Linux because their `.txt` fixtures are checked out as LF; building CRLF in-process gives Linux CI regression coverage for the specific bug.
- A new CRLF parity sub-step inside step 4 of `cargo xtask verify`. After the LF corpus passes, it copies the grammar to a tempdir, converts every `test/corpus/**/*.txt` to CRLF, and re-runs `tree-sitter test`. Locks in full-corpus CRLF parity for any future grammar/scanner change. Skippable via `--skip-treesitter-crlf-tests`. 451/451 corpus tests pass under CRLF; reverting the scanner fix correctly drops it to 446/451.

Fixes #138

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6935,6 +6935,7 @@ dependencies = [
  "clap",
  "proc-macro2",
  "syn",
+ "tempfile",
  "walkdir",
 ]
 
--- a/claude-notes/plans/2026-04-29-tree-sitter-crlf-parity-design.md
+++ b/claude-notes/plans/2026-04-29-tree-sitter-crlf-parity-design.md
@@ -1,0 +1,325 @@
+# Tree-sitter Corpus CRLF Parity Check — Design
+
+## Context
+
+PR #139 (bd-0gsj) fixes a CRLF bug in `crates/tree-sitter-qmd/tree-sitter-markdown/src/scanner.c` where pipe tables absorbed following paragraphs when input used `\r\n` line endings. The fix is one character (`!=` → `==`) plus a Rust regression test (`pipe_table_crlf_matches_lf`) that derives CRLF in-process and asserts the parsed S-expression matches LF.
+
+Lead dev cscheid asked on the PR:
+
+> We have dedicated tree-sitter tests too, but since this is whitespace related, I imagine that we would need a new strategy for those. What do you think about something like a script that converts all unix line breaks to windows linebreaks in the tree-sitter test files and then runs again? I think I would like the parses and errors to stay exactly the same. The source locations would change, but the tree-sitter test suite isn't checking source locations.
+
+This design wires that strategy into the build.
+
+## Goal
+
+After `tree-sitter test` passes against the LF corpus, re-run the same suite against a CRLF-converted copy. Identical parses, identical errors. Source locations differ, but `tree-sitter test` does not compare them — corpus expectations are bare S-expressions.
+
+This locks in the bug-fix and prevents any future grammar/scanner change from silently regressing CRLF behavior on Linux CI (where corpus files are checked out as LF).
+
+## Non-goals
+
+- Adding a Windows CI runner. Orthogonal workstream.
+- Removing `eol=lf` from `crates/tree-sitter-doctemplate/grammar/.gitattributes`. Deferred to the Windows CI workstream.
+- Covering the doctemplate grammar's corpus. Its scanner has no `\r` handling because newlines are not scanner-relevant in that grammar; CRLF risk is negligible.
+
+## Mechanism
+
+A new step in `cargo xtask verify`, sequenced after step 4 ("tree-sitter test"). Call it step 4b: "tree-sitter test (CRLF parity)".
+
+1. Create a temp directory.
+2. Mirror the grammar tree at `crates/tree-sitter-qmd/tree-sitter-markdown/` into the temp dir such that `tree-sitter test` invoked from the temp grammar dir picks up a CRLF copy of `test/corpus/`. Two viable mechanics:
+   - **(a) Symlink/junction the grammar dir, then physically copy `test/corpus/` with conversion.** Lightweight. Windows: junction (`mklink /J`) works for directory links without admin privileges.
+   - **(b) Full copy of the grammar dir.** Heavier (includes generated `src/parser.c`, bindings) but mechanically simpler.
+
+   Implementation plan picks one based on what works cleanly cross-platform.
+3. For each `*.txt` file in `test/corpus/`, normalize line endings to LF first (idempotency: handle files already CRLF), then replace LF with CRLF. The whole file is converted, including the `===`/`---` separators and the expected S-expression block — `tree-sitter test`'s parser of the corpus format is line-oriented and tolerates either ending.
+4. Invoke `tree-sitter test` with the temp grammar dir as cwd.
+5. Surface the same exit code as step 4. Skippable via a new `--skip-treesitter-crlf-tests` flag, mirroring the existing `--skip-treesitter-tests`.
+
+## Failure modes and policy
+
+Most likely outcome: all tests pass. The bug-fix in `scanner.c` was the only known CRLF-sensitive site; an audit during the fix found `\r` paired with `\n` everywhere else.
+
+If a corpus test fails under CRLF:
+
+- **Real grammar bug** — fix it, same as the pipe-table fix. Add a Rust regression test alongside `pipe_table_crlf_matches_lf` if the failing test isn't already covered by one.
+- **Test legitimately depends on bare-LF semantics** — not expected for any markdown construct, but possible. Last resort: exclude that single corpus test from the CRLF run and document why in this file. Decision deferred until a concrete case appears.
+
+## Why a script and not a Rust test
+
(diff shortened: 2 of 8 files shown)


cscheid:
Thank you, that's a really good catch.

We have dedicated tree-sitter tests too, but since this is whitespace related, I imagine that we would need a new strategy for those.

What do you think about something like a script that converts all unix line breaks to windows linebreaks in the tree-sitter test files and then runs again? I think I would like the parses and errors to stay exactly the same. The source locations would change, but the tree-sitter test suite isn't checking source locations.

cderv:
Thanks for the feedback. I'll look at this option.

cderv:
Done in 8f61f587e03569758a0308e0f3ceeee7bd4bf88f.

Step 4 of `cargo xtask verify` now re-runs `tree-sitter test` against a temp copy of the grammar with the corpus converted to CRLF. Skippable via `--skip-treesitter-crlf-tests`.

https://github.com/quarto-dev/q2/blob/8f61f587e03569758a0308e0f3ceeee7bd4bf88f/crates/xtask/src/treesitter_crlf.rs

451/451 corpus tests pass under CRLF on the fixed scanner. Reverting the one-character scanner fix drops it to 446/451 and fails the verify, so the parity check does catch the regression it's meant to.

Scope is the markdown grammar only — no CI workflow changes here. If a corpus test ever turns out to legitimately depend on bare-LF semantics, we could exclude it case-by-case; none of the current 451 do.

## About doctemplate and `.gitattributes`

The doctemplate grammar has `* text=auto eol=lf` in its `.gitattributes`, which forces LF on checkout everywhere, including Windows. That's why a Windows checkout doesn't naturally exercise CRLF for that grammar today — git rewrites the corpus to LF before tree-sitter sees it.

Two ways CRLF coverage could come in for doctemplate later:

- Drop `eol=lf` and let `text=auto` apply normal Windows CRLF conversion on checkout. Once we have Windows CI running `tree-sitter test`, that path would catch CRLF regressions there for free.
- Or: extend the parity runner in this PR to the doctemplate grammar too. Same mechanism, different grammar dir.

Not doing either here — doctemplate's scanner has no `\r` handling because newlines aren't scanner-relevant in that grammar, so the CRLF risk is negligible. Flagging it so it's not lost.

cderv:
@cscheid just asking for review so that you can verify I got your idea right.

cscheid:
Yeah, I think that's exactly what I was thinking, cool!!

cscheid (approved):
Excellent. This kind of automation really will reduce our regression surface on multiple OSs. Thank you!

PR #141, opened 2026-04-29
Reduce test-suite CI disk usage with profile.ci and runner cleanup

cderv:
The test-suite workflow on ubuntu-latest hit "No space left on device" during `cargo nextest run`, even with the existing free-disk-space step already reclaiming 18.8 GB. The workspace has grown past PR #55's mitigation.

## Root Cause

Two compounding pressures:

1. The default `[profile.dev]` emits full debuginfo (`debug = 2`), which roughly doubles `target/` size on a workspace this big. Dependencies dominate `target/` and carry no useful debug info for triaging q2 failures.

2. The workflow had both `cargo build` and `cargo nextest run` steps. Library crates compiled with `--cfg test` have a different fingerprint than dev builds and produce separate `.rlib` artifacts in `target/<profile>/`. With ~35 crates that runs into multi-GB duplication at peak disk.

## Fix

Four independent levers, applied in one workflow change:

- Add `[profile.ci]` to the workspace `Cargo.toml` (inherits from `dev`, `debug = "line-tables-only"` on the workspace, `debug = false` on all dependencies). Local builds keep full debuginfo via the unchanged `dev` profile.
- Replace `cargo build` + `cargo nextest run` with a single `cargo nextest run --tests --cargo-profile ci`. Per the [nextest design docs](https://nexte.st/docs/design/how-it-works/), nextest delegates compilation to `cargo test --no-run`, which makes the separate `cargo build` step redundant: the lib + bins + integration-tests that `--tests` selects strictly supersede `cargo build`'s default targets (lib + bins, non-test). `--tests` rather than `--all-targets` because `quarto-yaml`'s `harness = false` benches print prose reports that nextest can't enumerate as tests; the prior workflow never built benches anyway, so coverage matches. `RUSTFLAGS=-D warnings` continues to apply (it's a rustc env var, picked up regardless of which cargo subcommand drives the build).
- Enable `remove_tool_cache: true` on the existing `endersonmenezes/free-disk-space` step (~6 GB more). Verified safe: pre-installed rustup data lives in `/etc/skel/.{rustup,cargo}` and `dtolnay/rust-toolchain` installs to `~/.rustup/toolchains/`, neither of which is in `/opt/hostedtoolcache/`.
- Add `docker image prune --all --force` + `docker builder prune -a --force` (~3-8 GB more). Pre-pulled docker images aren't used by this job.

Full rationale, measured numbers, and held-in-reserve options (swap removal, df -h diagnostic, larger runner) are in `claude-notes/2026-04-28-ci-disk-space-and-profile-ci.md`.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -114,22 +114,16 @@ jobs:
         shell: bash
         run: cargo xtask lint
 
-      # Build and test
-      # Deny warnings in CI so they don't accumulate silently
-      - name: Build
-        shell: bash
-        run: cargo build
-        env:
-          RUSTFLAGS: "-D warnings"
-
       - name: Test block tree-sitter grammar
         shell: bash
         run: |
           cd crates/tree-sitter-qmd/tree-sitter-markdown
           tree-sitter test
 
+      # `ci` profile + `--all-targets` makes a separate `cargo build` redundant.
+      # See claude-notes/2026-04-28-ci-disk-space-and-profile-ci.md.
       - name: Test Rust code
         shell: bash
-        run: cargo nextest run
+        run: cargo nextest run --all-targets --cargo-profile ci
         env:
           RUSTFLAGS: "-D warnings"
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -100,7 +100,9 @@ jobs:
         if: runner.os == 'macOS'
         run: brew install tree-sitter-cli
 
-      # Free disk space on Linux runners (14 GB SSD is tight for Rust monorepo)
+      # Free disk space on Linux runners (14 GB SSD is tight for Rust monorepo).
+      # `remove_tool_cache: true` is safe — no step in this job uses /opt/hostedtoolcache/
+      # (no setup-node, setup-python, etc.). See claude-notes/2026-04-28-ci-disk-space-and-profile-ci.md.
       - name: Free disk space
         if: runner.os == 'Linux'
         uses: endersonmenezes/free-disk-space@7901478139cff6e9d44df5972fd8ab8fcade4db1 # v3
@@ -108,6 +110,14 @@ jobs:
           remove_android: true
           remove_dotnet: true
           remove_haskell: true
+          remove_tool_cache: true
+
+      - name: Prune Docker images
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          sudo docker image prune --all --force
+          sudo docker builder prune -a --force
 
       # Custom lint checks (before build to fail fast)
       - name: Run custom lints
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -130,10 +130,12 @@ jobs:
           cd crates/tree-sitter-qmd/tree-sitter-markdown
           tree-sitter test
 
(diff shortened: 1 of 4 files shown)


cderv:
Follow-up fix: `quarto-lsp` integration test was hardcoding `target/debug/q2`, which broke under `--cargo-profile ci` (binary now lands at `target/ci/q2`).

**Approach chosen:** derive the profile directory from `std::env::current_exe()` — the test binary itself sits at `target/<profile>/deps/`, so backing up two parents lands in the same `target/<profile>/` where the production bin is built. Profile-correct on all platforms with no env coupling.

**Why not `CARGO_BIN_EXE_<name>`:** that env var is package-scoped — Cargo only sets it for integration tests in the same package as the bin. Our LSP test lives in `quarto-lsp` while the bin is in `quarto`, so it's never set here.

**Why not `assert_cmd`:** `assert_cmd::cargo::cargo_bin("q2")` does exactly the same thing internally and is the idiomatic crate-based answer. We deferred adding it because this test doesn't use any of `assert_cmd`'s actual value — no `.assert()`, no stdout matchers, just JSON-RPC over stdio. The trigger to switch is the first test that wants `cargo run --` style command-driving with output assertions; at that point `cargo_bin()` is the standard binary-discovery helper that comes with the crate.

Full rationale in `claude-notes/2026-04-28-ci-disk-space-and-profile-ci.md` (Change 3 section).

cderv:
Following our discussion about whether dropping `cargo build` could miss build issues that `cargo nextest` doesn't catch, I went back and checked. I think we're safe.

The `cargo build` step has been in the workflow since it was first added, paired with `cargo test` from the start, with no rationale in the commit message or as a YAML comment:

https://github.com/quarto-dev/q2/blob/a4f16c90a3561ba11e029fbeb84fc44c552975fe/.github/workflows/test-suite.yml#L26-L32

That looks to me like the standard idiomatic Rust CI pattern (`build` then `test`) rather than a deliberate "catch what tests don't" decision.

On compile coverage, here's what each command builds by default:

| Command | lib (non-test) | lib (test cfg) | bins (non-test) | bins (test cfg) | integration tests | examples | benches |
|---|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| `cargo build` | ✓ | — | ✓ | — | — | — | — |
| `cargo nextest run` (no flag) | ✓ (as dep) | ✓ | ✓ | ✓ | ✓ | — | — |
| `cargo nextest run --tests` | ✓ (as dep) | ✓ | ✓ | ✓ | ✓ | — | — |
| `cargo nextest run --all-targets` | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |

`cargo nextest run` (any flag) is a strict superset of `cargo build`'s defaults. The `--tests` flag I landed on is a *narrowing* choice vs `--all-targets` — picked to skip `quarto-yaml`'s `harness = false` benches that nextest can't enumerate as tests. It's not what makes `cargo build` redundant; the redundancy holds regardless of which nextest target flag we use.

Only theoretical gap I can think of would be a target with `test = false` in its `Cargo.toml`. The repo has one (`pampa-fuzz`) but it's already excluded from the workspace, so neither path was building it.

References for the claims above:
- nextest delegates compilation to `cargo test --no-run`: https://nexte.st/docs/design/how-it-works/
- `cargo test` default target selection: https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection

PR #151, opened 2026-05-01
Add basic offline support with a service worker

vezwork:
Adds basic offline support using a service worker that caches everything that is loaded when online. You can e.g. refresh your browser tab even without network/server connection and the app will still run. Also, clicking between the project screen and editor screen is instant feeling.

looking in the storage tab of my browser, it looks like about 57mb of stuff gets cached.

Disclosure: Claude wrote all the changes in this PR.

#### Limitations:

- Only caches files once they've "naturally" loaded. This means the editor files won't be cached for offline use if you never visited the editor page (you stayed on the project screen) while online.
- Seems to require a build, not currently working on `npm run dev:fresh` for example, so not quite as easy to test with.
- **I am sure this would lead to significant issues with application versioning in practice**... since everything is cached, we need to know when to replace cached files with their new versions and stuff. I don't know if this is somewhat handled by the current approach or not.
- This does not attempt to address the problem of intercepting HTTP requests and fulfilling them using stuff from the Quartohub Virtual File System (VFS) (e.g. for images stored in the VFS that are used in the preview), which is another thing we might want a service worker for.


cscheid:
> I am sure this would lead to significant issues with application versioning in practice... since everything is cached, we need to know when to replace cached files with their new versions and stuff. I don't know if this is somewhat handled by the current approach or not.

I _think_ the vite deployment strategy should work here (vite uses hashes in the filename of its assets).

cscheid:
@vezwork Is this something we can eventually merge or have we done the work in a different branch?

vezwork:
@cscheid yea I want to merge this, gonna work on it today

vezwork:
@cscheid @shikokuchuo do y'all know, will the PWA that `npm run build` produces be what is deployed? Or do we have to modify the deployment to use that? Otherwise this should be ready to try out, gonna merge soon.

PR #172, opened 2026-05-11
ci: get Hub-Client E2E Playwright Tests running

gordonwoodhull:
1. Get E2E Playwright tests running.
2. Enable them for PRs.

## Summary

The `Hub-Client E2E Tests` workflow had never actually executed on a runner since it was added — every push to `main` queued for the 24-hour Actions maximum and was cancelled because `runs-on: ubuntu-latest-8x` isn't a runner label this repo can request. This PR gets the workflow running and green on a stock `ubuntu-latest` runner.

Three commits, deliberately split by concern:

- **`ci: enable Hub-Client E2E Tests workflow`** — every fix needed in `.github/workflows/hub-client-e2e.yml`: runner label, build ordering, WASM tooling (wasm-bindgen-cli pinned to Cargo.lock, rust-src component, nightly pin to dodge a rustc SIGSEGV on tokio for wasm32), dtolnay action ref form, TS build scoping, hub binary pre-build, and a fix for the broken baseline auto-commit step (glob + permissions).
- **`test(e2e): make hub-client e2e suite pass on CI`** — every fix needed in the test code that the workflow surfaced once it could actually run: a TS2345 in the vitest mock helper, an `indexedDB` shim in the Playwright Node-side helper, a Vite proxy env to point `/auth/*` at the e2e hub's port, a `testIgnore` for visual specs so they only run via the dedicated config, cutting CI workers from 4 → 2 to match the 2-core runner, and a `SKIP_WASM_UNSUPPORTED` entry for the one fixture that exercises a real WASM gap.
- **`Add missing Playwright visual regression baselines`** — 6 auto-generated chromium-linux baselines committed back by the workflow itself on its first successful run; kept as a separate commit by `github-actions[bot]` to preserve provenance.

The remaining WASM gap is **bd-izfv** (Phase 9 follow-up: thread `user_grammars` through `RenderToHtmlRenderer`). A complete TDD plan for that fix is already on branch `beads/bd-izfv-thread-user-grammars`. When bd-izfv lands, the `SKIP_WASM_UNSUPPORTED` entry added in this PR can be removed and the user-grammar fixture will run; Phase 5 of that plan explicitly calls out the restore step.

Filed during this work:
- bd-1jnb (P2): `q2-demos/*` vite build fails resolving `/src/wasm-js-bridge/cache.js` — the workflow scopes the TS build to skip them.
- bd-233j (P3): noisy `[TAG_RESOLVE_FAILED] YAMLWarning` from `!str` tags in playwright stdout — cosmetic, not a failure cause.

## Test plan

- [x] Successful CI run: https://github.com/quarto-dev/q2/actions/runs/25645506576 (conclusion `success`, 76 tests passed, 6 baselines auto-committed)
- [x] Local repro of the workers-contention failure (4 workers reproduced random preview-iframe timeouts on macOS; dropping to 2 cleared them)
- [x] Single-test local pass via `npx playwright test --grep "01-builtin-python"` in the worktree (5.7s)
- [x] `git diff origin/main HEAD` reviewed and the 3-commit history is lossless against the pre-squash branch


gordonwoodhull:
Two additions after the initial PR:

- **`pull_request` trigger on `hub-client-e2e.yml`** (folded into the first commit). PRs that touch `hub-client/**` or the workflow file now run the e2e suite automatically, so this very PR will prove itself green before merge. Same path filter as the existing `push: branches: [main]` trigger.

- **Cherry-picked `ts-test-suite.yml` nightly pin** (new commit `af0eec95`, originally `aa3e9ad8` on `feature/q2-preview`). The pre-existing `TS Test Suite` workflow was failing on this PR with the same rustc-SIGSEGV-on-tokio-wasm that we already worked around for the e2e workflow; picking up the same fix keeps the matrix legs green. When `feature/q2-preview` lands first, this commit will deduplicate out (or be dropped during the eventual rebase).

The branch is now four commits, with all `.github/workflows/` changes grouped at the start of the stack and the bot's baselines commit kept separate at the tip.

PR #207, opened 2026-05-15
feat(pampa): better diagnostic for line breaks in inline link destinations

rundel:
When an inline link destination contains a hard line break (e.g. `[text](https://\nhost/path)`), pampa rejected the input with a generic "unexpected character or token here" message. CommonMark forbids line breaks in inline link destinations, so the rejection is correct, but the message is unhelpful.

This adds a Q-2-37 corpus entry covering the (LR state, lookahead) pairs the parser hits when a newline appears inside a link destination (state 2619 with both `shortcode_name` and `_close_block` lookaheads). Pampa now emits a specific message naming the cause and a single hint.

Output:

```
  Error: [Q-2-37] Line break in link destination
     ╭─[ file.qmd:2:1 ]
     │
   2 │ allisonhorst.github.io/palmerpenguins/) dataset.
     │ ──────┬─────
     │       ╰─────── An inline link destination cannot contain a line break.
  ───╯
  ℹ Remove the line break inside the link's `(...)` destination and keep the URL on a single line.
```

A new regression test (`tests/test_link_destination_linebreak.rs`) covers the bug-report fixture plus three minimal variants (newline after the scheme, mid-destination, before the closing `)`). The existing `test_error_corpus_*` tests automatically exercise the three generated case files.

The error location seems not ideal but I was not able to fix that without subjecting the codebase to much more significant changes than seemed warranted so I will leave that up to you to address.

### DIFF
--- a/crates/pampa/resources/error-corpus/Q-2-37.json
+++ b/crates/pampa/resources/error-corpus/Q-2-37.json
@@ -1,0 +1,29 @@
+{
+  "code": "Q-2-37",
+  "title": "Line break in link destination",
+  "message": "An inline link destination cannot contain a line break.",
+  "notes": [],
+  "hints": [
+    "Remove the line break inside the link's `(...)` destination and keep the URL on a single line."
+  ],
+  "cases": [
+    {
+      "name": "after-scheme",
+      "description": "Newline immediately after the scheme `://` of an inline link destination.",
+      "content": "[text](https://\nexample.com/path)\n",
+      "captures": []
+    },
+    {
+      "name": "mid-destination",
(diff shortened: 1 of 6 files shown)


rundel:
Also TIL about the Jeffery TOPLAS-2003 technique - super cool

cscheid:
> Also TIL about the Jeffery TOPLAS-2003 technique - super cool

Yes, it truly is! I only know about it because of a PL prof friend of mine. We're currently running into the limits of the technique for error messages here because parsing Markdown is, well, hell. But it's super cool, super simple, and wild that we could make it work inside tree-sitter.

PR #225, opened 2026-05-21
perf(attribution): speed up the Authors overlay producer (~30× at 10k changes)

shikokuchuo:
## What

Four commits, each landing a discrete win on the JS-side attribution producer (`buildRunListAttribution` / `useAttribution`, the code that builds the payload behind the Authors overlay).

1. **Replay history via `A.applyChanges` instead of per-step `A.diff`** (`7a6bc8a4`). The old loop called `A.diff(doc, prevHeads, currHeads)` per history step, which scales super-linearly with doc state. The new loop pre-indexes changes by hash and forward-replays through `applyChanges` + `patchCallback`, which is O(1) per change.

   | N      | diff() | applyChanges | speedup |
   |--------|--------|--------------|---------|
   | 500    | 22 ms  | 8 ms         | 2.6×    |
   | 2 000  | 197 ms | 22 ms        | 9.0×    |
   | 10 000 | 4 121 ms | 116 ms     | 35.5×   |
   | 30 000 | 36 325 ms | 375 ms    | 96.8×   |

   Runs verified equivalent across append-only, mixed-edit, and multi-actor (sequential + diamond-DAG) fixtures.

2. **Drop one yield per build** (`abee1827`). Moved the chunk-yield from the top of the loop to between chunks; a history that fits in one chunk (≤ `CHUNK_SIZE`) now never pays an idle-callback round-trip and short builds finish synchronously.

3. **`Uint32Array` for the char→byte map** (`a35e33e0`). Rebuilt on every debounced payload; switching from boxed `Array<number>` to a single contiguous `Uint32Array` cuts allocations and halves the per-codeunit heap footprint. 5–30 % faster from ~10 KB documents up; smaller docs are at-worst-noise.

4. **`CHUNK_SIZE` doc comment** (`a5081334`) — refreshed to reflect the new applyChanges cost profile (≈15 µs/entry, roughly constant in N) instead of the obsolete `A.diff` numbers.

## Why

Cold-start of the Authors overlay on long histories was the dominant wallclock, driven by `A.diff` reloading historical doc state on every step. The rewrite makes it linear; the smaller wins remove the leftover papercuts so the producer is fast even on a 30 k-change document (375 ms vs the old 36 s).

### DIFF
--- a/hub-client/src/services/attribution-runs.test.ts
+++ b/hub-client/src/services/attribution-runs.test.ts
@@ -29,13 +29,13 @@ describe('buildCharToByteMap', () => {
     // "é" is U+00E9, 2 bytes in UTF-8 (0xc3 0xa9).
     const map = buildCharToByteMap('aéb');
     // 'a' at char 0 → byte 0; 'é' at char 1 → byte 1; 'b' at char 2 → byte 3.
-    expect(map).toEqual([0, 1, 3, 4]);
+    expect(Array.from(map)).toEqual([0, 1, 3, 4]);
   });
 
   it('counts 3-byte UTF-8 sequences correctly (CJK)', () => {
     // "中" is U+4E2D, 3 bytes in UTF-8 (0xe4 0xb8 0xad).
     const map = buildCharToByteMap('a中b');
-    expect(map).toEqual([0, 1, 4, 5]);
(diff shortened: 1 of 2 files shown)


cscheid:
Excellent!!

PR #231, opened 2026-05-22
feature: improve provenance and make q2-preview editable

gordonwoodhull:
Draft PR for CI. 

Provenance epic plans 3-7 are complete; provenance data is flowing and impossible edits to atomic elements are both blocked on the front end and soft-dropped by the incremental writer.

Next up:
* Plans 7a-7c: more testing and a soft drop warning for attempting to edit before first render
* Plan 8: use a custom node for include shortcode
* Plan 9: expose provenance of YAML values used in transforms/meta shortcode
* Plan 10: consistent provenance of Lua-produced content, so that you could eg cmd-click some content and go to the line of Lua code that produced it.


gordonwoodhull:
This approach turned out to be way too complicated. It is archived in `archive/provenance-filtered-edits`

replaced by

* #260

PR #241, opened 2026-05-29
Mermaid diagram engine + RawBlock script re-execution for q2 preview parity

cscheid:
Adds a `{mermaid}` diagram engine to Quarto 2 and makes it render
identically under `q2 render` (static HTML) and `q2 preview` (live
React iframe). End-to-end verified in both paths; epic + sub-tasks
tracked under bd-je48v.

## What's in here

| Phase | What landed | Beads |
|---|---|---|
| Design + audit | Plan doc + design ratification + PR #238 audit | bd-c6h96, bd-fztki |
| Engine | `MermaidEngine` (text-level fence scanner, B1 direct HTML emission) + 13 unit tests + 4 integration tests | bd-gwfdo |
| Preview parity | RawBlock script-re-execution shim in `q2-preview` and slide renderers + 9 vitest tests; engine emits `mermaid.run()` instead of `startOnLoad:true` | bd-my0o5 |
| Closes | PR #238's bd-iq0hp (multi-engine browser e2e) | bd-iq0hp |

Plan + audit doc: `claude-notes/plans/2026-05-28-mermaidjs-engine-design.md`.

## The engine

```
```{mermaid}
graph TD
A --> B
```
```

Becomes:

```html
<pre class=\"mermaid\">graph TD
A --&gt; B
</pre>
<script type=\"module\">
import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs';
mermaid.initialize({ startOnLoad: false });
mermaid.run({ querySelector: 'pre.mermaid' });
</script>
```

Mechanics:

- **Engine model:** plugs in through PR #238's `engine: [a, b, ...]`
  sequence model — `engine: mermaidjs` (scalar) or
  `engine: [knitr, mermaidjs]` (array) both work. Subprocess-free,
  registers in both native and WASM `EngineRegistry`s, so the
  preview's WASM build runs mermaid live in the browser without
  needing a recorded capture round trip.
- **Text-level scanner**, mirroring `FixtureEngine`'s precedent — a
  hand-rolled fence parser, no pampa AST involvement.
- **B1 emission** (direct raw HTML). Format-locked but acceptable
  while Q2 is HTML-only; bd-mqk49 (engine → stage extension API)
  tracks the eventual refactor to a marker `Div` + format-conditional
  AST pass. A source-code comment at the emission site points at
  bd-mqk49.
- **Pandoc raw-block fence (` \`\`\`{=html} `):** required because pampa's
  QMD reader treats bare `<tag>` markup at block position as
  `RawInline` and tries to parse the interior as Markdown — which
  breaks on `startOnLoad: true` (the `:` reads as a definition list).
  The explicit fence form keeps the contents opaque to the reader.

## The preview parity work

`q2 render` worked from the first commit. `q2 preview` exposed two
compounding gaps that both had to be fixed for the diagrams to
render in the React iframe — they are documented in detail in the
plan doc's Phase 3 section and at length in the code comments.

### P1: `dangerouslySetInnerHTML` does not execute inline `<script>`

The q2-preview format's `RawBlock(html)` React component injects
raw HTML via `dangerouslySetInnerHTML`, which sets `innerHTML`,
which invokes the HTML *fragment* parsing algorithm. The fragment
parser flags every `<script>` it creates as **\"already started\"**
(HTML Living Standard § 4.12.1) so the spec's \"prepare the script\"
algorithm returns early without evaluating the body. The element
sits in the DOM and is queryable; the body never runs.

**Fix** (`ts-packages/preview-renderer/src/q2-preview/blocks/RawBlock.tsx`):
after mount, walk the container for `<script>` elements and replace
each with a freshly created element via
`document.createElement('script')` carrying the same attributes and
body. The replacement is a \"script-created\" element in HTML-spec
terms — \"already started\" is not set — so the browser runs it.

**Security implications**, ratified with @cscheid 2026-05-29: scripts
inside `RawBlock(html)` will now execute in the preview iframe, the
same posture `q2 render`'s static HTML already has. This is a
behaviour change for the preview pipeline which previously silently
ignored such scripts. The PR is intentionally on its own
`feature/mermaid-engine` branch so the hub-client team can review
the security model change before this lands on `main`.

The same shim lands on the slide renderer's parallel RawBlock case
for behavioural parity, even though slides aren't currently the
mermaid-exercising path.

### P2: `mermaid.initialize({startOnLoad: true})` no-ops on dynamic injection

`startOnLoad` only fires when `mermaid.initialize` is called *before*
`DOMContentLoaded`. In static render the script ships in the initial
document, so the listener fires. In preview the script reaches the
iframe long after the document has loaded (the shim above re-creates
the script element so it executes), and `startOnLoad` silently
no-ops.

**Fix** (`crates/quarto-core/src/engine/mermaid.rs`): emit
`mermaid.run({ querySelector: 'pre.mermaid' })` after
`mermaid.initialize({ startOnLoad: false })`. `mermaid.run` works
regardless of load timing and is idempotent on already-processed
elements (mermaid tracks via `data-processed`).

## Tests

- 13 unit tests in `engine/mermaid.rs` (single/multi/no-cell, mixed
  with other engines, HTML escaping, longer fences, unterminated
  cells, raw-HTML-fence structural properties).
- 4 integration tests in `crates/quarto-core/tests/mermaid_pipeline.rs`
  routed through `render_to_file` — the same path `q2 render` uses
  (single-doc emits pre+script, no-cells omits script, multi-cell
  shares one script, array engine form `[mermaidjs]` works).
- 4 vitest tests in `ts-packages/preview-renderer/src/q2-preview/blocks/RawBlock.test.tsx`
  covering the React shim. jsdom doesn't execute scripts by default,
  so these assert the structural recreation; real execution is
  proven by the browser run below.
- 3 parallel vitest tests on the slide renderer's RawBlock case.
- Full `cargo nextest run --workspace`: 9501 tests pass (no
  regressions from the registry / `KNOWN_ENGINES` additions).
- Full hub-client `npm run test:ci`: 79 tests pass (incl. 61 WASM
  smoke-all fixtures).
- `cargo xtask verify` (full, with WASM + hub-client build): all 12
  steps green.

## End-to-end verification (real binary, real browser)

Static `q2 render` (full record in plan doc):

```
cargo run --bin q2 -- render /tmp/mermaid-e2e/test.qmd
```

Rendered HTML contains both `<pre class=\"mermaid\">…</pre>`
wrappers + the jsdelivr `<script type=\"module\">` include.

`q2 preview` (Chrome DevTools MCP):

```
cargo run --bin q2 -- preview /tmp/mermaid-preview/
# → opens http://127.0.0.1:<random>/
```

Iframe inspection after 5 s settle:

```js
{
  preCount: 2,
  preProcessed: ['true', 'true'],   // mermaid marks rendered elements
  svgInsidePre: 2,                  // both diagrams rendered as SVG
  recreatedScriptCount: 1,          // RawHtmlBlock shim swapped the inert <script>
}
```

Console clean. Screenshots before/after both fixes:
`claude-notes/plans/bd-my0o5-preview-{no-render,renders}.png`.

## Honest gaps + follow-ups

- **Multi-engine integration test** (`engine: [knitr, mermaidjs]`
  with one R cell and one mermaid cell). Deferred — gated on knitr R
  runtime availability or a FixtureEngine adapt. Single-engine
  mermaid + the array form `[mermaidjs]` are both covered; the
  multi-engine variant uses the same code path.
- **bd-mqk49** — engine → stage extension API. When that lands,
  refactor the mermaid engine to emit a marker `Div` and declare an
  HTML-conditional AST pass instead of inlining raw HTML. Tracked as
  a discovered-from on the design issue.
- **bd-cp3em** — `CaptureSpliceStage` still drops aux fields
  (`includes`/`filters`/`supporting_files`) from recorded
  `ExecuteResult`s. Mermaid sidesteps the bug because it registers
  in WASM directly and runs live in-browser — but the gap is real
  for any engine that emits side effects via `includes`.
- **bd-14rer** — `ExecuteResult.filters` written by Knitr but never
  consumed downstream (`rmarkdown/pagebreak.lua` silently dropped).
- **bd-s8llm** — no internal API for pipeline producers to publish
  files into the VFS.
- **bd-5ijtt** — user-facing docs page (intentionally deferred until
  after review).


cscheid:
We're going to wait to merge this one until we have a good design for iframe security/scripts.

PR #246, opened 2026-05-31
fix(pampa): detect parse errors without the per-lex tree-sitter logger (bd-b7eb7)

cscheid:
## Summary

The qmd reader (`crates/pampa/src/readers/qmd.rs::read`) attached a tree-sitter `set_logger` callback on **every** parse. When *any* logger is set, tree-sitter formats a debug string with `snprintf` for every lex action — and on macOS `snprintf` locks the global locale (`localeconv_l`). Under multithreaded project renders (rayon, ~16 workers) that locale lock serializes all parser threads.

## How it was found

Profiled a large **healthy** project (401 docs / 8.2 MB, no errors) — the case we most want fast. A fresh multithreaded render spent **~75% of samples** in `__ulock_wait2`/`__ulock_wake`/`os_unfair_lock`, call chain:

```
ts_lexer__advance → snprintf → __vfprintf → localeconv_l → os_unfair_lock
```

16 threads bought only ~1.2× wall-clock; `sys` time ~37 s. It is **not** the allocator — swapping mimalloc both as `#[global_allocator]` *and* via `tree_sitter::set_allocator` changed nothing. The locale-lock culprit was found with macOS `sample` (which symbolicates system libraries, unlike samply's sidecar).

## Fix

Parse the happy path with **no logger**, and detect whether tree-sitter entered error-correction mode via its own public signal — the `parse_with_options` progress callback's **`ParseState::has_error`** — combined with the final tree's `Node::has_error` (which backstops an error in the last few ops before parse end). The logger is attached only for the diagnostic **re-parse**, which runs only when the document has an error. Encapsulated in `MarkdownTree::had_parse_error`. The previous, ineffective `TreeSitterLogObserverFast` (its cheap callback didn't help — the `snprintf` is inside tree-sitter, before the callback) is removed.

**Why `has_error` is reliable here:** tree-sitter-qmd declares **no `conflicts:`**, so the parser is deterministic LR and GLR multiple-stack speculation only occurs during error recovery — `has_error` therefore means a genuine error, not a speculative branch that will be pruned. This invariant is now documented in `grammar.js` and `parser.rs`.

## Results (large healthy project, `release-perf`)

| metric | before | after |
| --- | --- | --- |
| 16-thread render | 8.04 s | **3.71 s** (2.2×) |
| 16-thread `sys` | 36.83 s | **1.37 s** (27×) |
| pass-1 @ 16 threads | 2935 ms | **426 ms** (scales ~6.4× on 16 cores) |
| single-thread render | 9.71 s | **6.0 s** (1.6×) |

## Correctness

- Error-corpus snapshot tests (json / text / ariadne / locations) pass.
- The real 565-file qmd-plans corpus reports the **same 140 error files** — exact parity with the previous log-based detection.
- Full `cargo xtask verify` (Rust workspace + WASM/hub build + hub tests) passes.

## Notes

- Also documents the profiling lesson (use macOS `sample` for system-lib frames; a Rust `#[global_allocator]` doesn't cover a vendored C library's libc `malloc`) in `claude-notes/instructions/performance-profiling.md`.
- Full investigation: `claude-notes/research/2026-05-31-q2-render-perf-qmd-plans.md` (lands separately with the website experiment).

### DIFF
--- a/claude-notes/instructions/performance-profiling.md
+++ b/claude-notes/instructions/performance-profiling.md
@@ -159,6 +159,45 @@ top-N self-time table. The head of the script documents both the
 profile format and the samply sidecar structure — if samply's format
 changes, patch there.
 
+#### When the hotspot is in a system library, use `sample`
+
+samply's `.syms.json` sidecar symbolicates **our** binary, but it does
+**not** resolve system libraries (`libsystem_c.dylib`,
+`libsystem_malloc.dylib`, …). Frames inside them show up as bare
+addresses (`0x4a83 [libsystem_c.dylib]`). If a hotspot — or, in a
+multithreaded profile, the frame *holding a contended lock* — lands in a
+system library, the address tells you nothing, and **guessing what it is
+will burn your session.**
+
+macOS ships `sample`, which symbolicates system libraries from the dyld
+shared cache:
+
+```bash
+target/release-perf/<binary> <args> &   # or run it and grab the pid
+sample $! 4 -file /tmp/s.txt            # sample the live process for 4s
+# then read /tmp/s.txt — system frames now have names
+```
+
+Real example (bd-b7eb7): a multithreaded `q2 render` showed ~75 %
+`__ulock_wait2`/`os_unfair_lock` with the waits charged to tree-sitter's
+`ts_lexer__advance`, and the libc frame below it unsymbolicated. We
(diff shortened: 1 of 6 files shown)


cscheid:
Superseded by #247 — same commits, rebranched to `perf/tree-sitter-error-logging`.

PR #260, opened 2026-06-05
feat(provenance): targeted paragraph/header inline editing

gordonwoodhull:
Editing of most blocks work - a couple of quirks I'll point out when I demo, but I think this is very close to merging.

There is one plan (2b) with looks ends around keyboard and touch, which I hope to implement.

There's a fourth plan in the epic which allows editing of sections. After looking at the results of plan 3, I'm not sure this is necessary.

This uses a variant of @vezwork's idea for targeted edits, modified for the pipeline-transformed AST. Specifically it uses a trick devised by @cscheid:
1. Pass both the untransformed (parsed) AST, and the pipeline-transformed AST to q2-preview.
2. `setAst` takes a source location and pure untransformed AST for the replacement, not the same data it draws.
3. We splice in the new AST in the old location and run reconciliation/coarsen/apply changes as before.

Currently the untransformed AST is only used to find the source location, but I think it will be helpful for render components that want to do a transformation. Currently kanban and comments don't work but they should be able to use this data.

Needs review and cleanup esp. with regard to DX around the two ASTs. It will make render components slightly more complicated but it's a good clean model overall.

One surprise: adding markdown inside the contentEditable does work - even block stuff like adding a bullet list etc. OTOH other markdown formatting in the same paragraph gets lost due to the contentEditable round-trip. @cscheid's suggestion to show the markdown in the contenteditable instead of just text should fix the latter.

## Summary

- Adds `apply_node_edit` — a WASM entry point that splices a user-edited AST subtree back into the source QMD using source-range provenance, producing a minimally-diffed rewrite.
- Wires an inline edit surface into `q2-preview`: clicking a `Para` or `Header` block activates `contentEditable`; blur/Enter commits the edit back through `apply_node_edit`; Escape cancels.
- Fixes a React reconciliation bug where the browser-mutated contentEditable DOM was left in place after commit because React's fiber reconciler compared against its last-committed virtual DOM (not the actual DOM). Fix: increment a `viewKey` counter only on deactivation, forcing a clean remount of the element after each edit without triggering the nbsp-conversion side-effect that a symmetric activation remount caused.

## Test plan

- [ ] Start hub dev server and open a `.qmd` file in q2-preview mode
- [ ] Click a paragraph — confirm blue outline appears, text is editable
- [ ] Type changes, blur or press Enter — confirm text updates in preview and Monaco editor shows clean spaces (no ` ` yellow boxes)
- [ ] Press Escape — confirm edit is discarded and original text is restored
- [ ] Edit a heading (`h1`–`h6`) — same behaviour
- [ ] Edit multiple paragraphs in sequence — confirm no duplicated/concatenated content glitch
- [ ] Playwright e2e: `npm run test:e2e` from `hub-client/`


ts-packages/preview-renderer/src/q2-preview/blocks/Para.tsx
```diff
@@ -1,4 +1,62 @@
...
+
+    const deactivate = (commit: boolean) => {
+        if (commit) {
+            const el = textRef.current;
+            if (!el) return;
+            // Normalize   (non-breaking space) that contentEditable inserts
+            // at text boundaries to prevent whitespace collapsing.
+            const newText = el.innerText.trim().replace(/ /g, ' ');
+            if (newText) {
+                ctx!.commitEdit!(poolId!, newText + '\n');
```
vezwork: Does `ctx.commitEdit` take text, parse it to a sub-AST, find the node with given `poolId` and get its source location, then splice the `write`'d parsed sub-AST into that source location in the qmd string?

vezwork:
I want to understand how this untransformed AST (for source locations) + transformed AST (for rendering) approach works, here's a question to try to get into that: 

What happens in the case that a Para node is generated as part of a transformation step. Will it not have a `poolId` in the render component?

vezwork:
Is it that Paras and Headers are the only thing working because those are the only render components that have been modified to do `commitEdit`? Or is there something deeper in this approach that for now limits it to only some nodes?

gordonwoodhull:
> Is it that Paras and Headers are the only thing working because those are the only render components that have been modified to do `commitEdit`? Or is there something deeper in this approach that for now limits it to only some nodes?

The Para and Header restriction was for the purpose of having a quick demo. I'm currently extending this to all top-level blocks and I want to try inner blocks and also sections. 

I'll post an update today, and demo tomorrow.

Two goals:
1. Multiple levels of block edit => proves the "AST-targeted edits" algorithm.
2. An interface that can be made to work across mouse, keyboard, and touch => hopefully will not leave us underwater in technical/design debt if we decide to keep it.

I am not sure if it will be an elegant interface or the the one we want; glad to throw it out if we come up with something better.

gordonwoodhull:
> Does ctx.commitEdit take text, parse it to a sub-AST, find the node with given poolId and get its source location, then splice the write'd parsed sub-AST into that source location in the qmd string?

Here's a silly and sloppy but accurate depiction of the edit cycle - text generated in the iframe gets parsed by the parent (who has wasm) and then the back-end splices and runs the usual reconciliation/coarsen/write.

Technically we don't need tree reconciliation, but I think it's simpler still to use it unless we find a reason not to. 

Reconciliation quickly identifies exactly the splice and creates a plan including only that.

q2-preview-edit-cycle-v3

gordonwoodhull:
> I want to understand how this untransformed AST (for source locations) + transformed AST (for rendering) approach works, here's a question to try to get into that:
> 
> What happens in the case that a Para node is generated as part of a transformation step. Will it not have a `poolId` in the render component?

If the paragraph is generated it will either have no poolId (here, will not have `node.s` source info entry), or the pool entry `node.s` will have `t !== 0`, in which case it's not `Original` and can't be edited.

However, any inlines under the paragraph could be shortcodes or transformed, and that's fine, we don't care. They will just show as markdown in the editor.

gordonwoodhull:
Now partially squashed and cleaned up, all top-level blocks are editable, all render components are working.

See e2e tests for the new render component 2-AST architecture.

gordonwoodhull:
Despite some bugs, I think I want to merge this now, once it passes tests.

It's getting slow to rebase it, and the provenance work deprecates `SourceInfo::default()`...

Will follow up with further fixes.

PR #270, opened 2026-06-10
fix(bd-jv5u083l): renewal verdict deadline unwedges expiry logout

shikokuchuo:
## Background

How the hub client manages session expiry today (`useAuth`):

1. On page load, `GET /auth/me` checks whether the auth cookie is valid; a 401 shows the login screen.
2. Fifteen minutes before the token expires, the client asks Google One Tap for a fresh credential ("silent renewal") and exchanges it for a new cookie via `POST /auth/refresh`.
3. At the expiry time itself, the client re-checks `/auth/me`. A definitive 401 logs the user out (with a "session expired" message); a network error keeps the session and re-checks every 60 seconds — so offline editing survives expiry, and the dead session is adjudicated as soon as the hub is reachable again.
4. While a renewal is in flight, the client sets an `isRefreshing` flag so that concurrent triggers coalesce, the tab-refocus check stands down, and the expiry re-check waits for the renewal's verdict instead of logging the user out underneath it.

## The bug

Step 4 assumed the identity provider always answers. It doesn't: if the One Tap prompt never fires (script blocked by an ad blocker, FedCM / third-party-cookie policies, unregistered origin — we've observed `gsi/status` 403 in dev), neither the success nor the error callback ever runs, and `isRefreshing` stays true for the lifetime of the tab. Everything gated on that flag then silently dies:

- the at-expiry re-check loops forever "waiting for a verdict" — the user is never logged out at expiry;
- the refocus check early-returns — no recovery when the tab regains focus;
- every later `triggerRefresh()` is a no-op — no retry, ever.

So in exactly the environments where silent renewal can't work, the client neither renews nor logs out, until a full page reload or a websocket drop.

## The fix

A 30-second verdict deadline on the in-flight renewal. If the provider hasn't settled by then, the renewal counts as failed and the flag clears, so the expiry re-check reaches its verdict normally: 401 → logout, network error → keep waiting.

The regression test mounts a session, lets the renewal trigger, and never settles it; it fails on `main` (the session survives expiry) and passes with the fix.

### DIFF
--- a/hub-client/src/hooks/useAuth.ts
+++ b/hub-client/src/hooks/useAuth.ts
@@ -39,6 +39,9 @@ const DEFAULT_SESSION_MS = 3600 * 1000;
 /** Re-check interval when an expiry-time verdict couldn't be reached. */
 const EXPIRY_RECHECK_MS = 60 * 1000;
 
+/** Time the IdP gets to settle a renewal before it counts as failed (30 s). */
+export const REFRESH_VERDICT_TIMEOUT_MS = 30 * 1000;
+
 export function useAuth() {
   const provider = useAuthProvider();
   const [auth, setAuth] = useState<AuthState | null>(null);
@@ -48,6 +51,7 @@ export function useAuth() {
   const isRefreshing = useRef(false);
   const refreshTimer = useRef<ReturnType<typeof setTimeout>>(null);
   const expiryTimer = useRef<ReturnType<typeof setTimeout>>(null);
+  const refreshDeadline = useRef<ReturnType<typeof setTimeout>>(null);
 
   // Effective expiry of the current session (0 = no session).
   const expiresAtRef = useRef<number>(0);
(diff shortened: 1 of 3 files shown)


hub-client/src/hooks/useAuth.ts
```diff
@@ -100,14 +97,13 @@ export function useAuth() {
...
-            expireSession();
-          }
+          if (me) applyAuth(me);
+          // A renewal that yields no session is not logout evidence — the
+          // open WS may still be syncing. The probe escalates if the socket
+          // is actually rejected (bd-jv5u083l).
         })
         .catch(() => {
-          if (sessionLapsed()) expireSession();
+          // Network/server error during renewal — not logout evidence.
```
cscheid: Do I read this diff correctly that we now have a `.catch(() => {})` clause? Doesn't this eat any and every  exception raised?

shikokuchuo: I think this was indeed not correct - led me to re-think the changes here and I decided to narrow the scope of this PR.

PR #275, opened 2026-06-11
chore(e2e): Playwright 1.60, fix render-queue race, smoke-all opt-in

gordonwoodhull:
## Summary

- **Playwright 1.60.0**: bumps `@playwright/test` from 1.58 to 1.60 and restores `node-version: '24'` in CI. Playwright 1.60 contains the upstream fix (microsoft/playwright#40747) for the yauzl extraction hang that forced the Node 24.15.0 pin. Chromium upgrades 145→148; visual baselines need regeneration via `recreate-all-snapshots` after merge.

- **Fix render-queue race (root cause of most flakiness)**: `render_page_in_project_with_attribution` is a Rust `async fn` — it yields the JS event loop at internal await points. When Automerge delivers file updates in rapid succession, two `renderToHtml` calls can overlap, sharing the same WASM VFS in undefined ways and silently producing blank preview iframes. Fix: a promise-chain serial queue in `wasmRenderer.ts` ensures only one render runs at a time. Validated with parallel CI experiments — the queue alone is sufficient; `vfsClearAsync` in `connect()` was explored and reverted (it stalled test setup).

- **Restore sync-gate fix** (dfbc2711, lost in Phase 6 revert 83efdb99): `seedProjectInBrowser` wrote to IDB only; the app's reconciler never re-fired once already connected. Fix: expose the reconciler on `window.__quartoTest` and wait for the project to appear in the synced set. Refs bd-3nzyd.

- **Fix actor ID injection**: `page.addInitScript` set `__QUARTO_TEST_ACTOR_ID__` but nothing read it, so `useCurrentActor()` returned null in the preview iframe and the reactji remove-mine path never fired. Fix: `connectAndLoadContents` reads the override when `VITE_E2E=1` (tree-shaken in production). Comment fixture updated to use attribution-based remove instead of the fragile `mySessionReactions` ref.

- **Replace theme-subdir-e2e with static smoke-all fixtures**: the 9 Playwright theme/SCSS tests were redundant — theme resolution is a static render-time decision already exercised by `ensureCssRegexMatches` in the WASM and Rust smoke-all pipelines. Replaced with two new smoke-all fixtures covering the three-layer metadata cascade and subdir doc override scenarios.

- **Smoke-all opt-in**: default CI now runs 21 custom tests only (~10 min total, ~1.5 min test execution). Smoke-all (~80 tests) is available on demand via `workflow_dispatch` with `run-smoke-all=true` — useful after changes to the WASM pipeline or theme resolution. Smoke-all remains valuable for the full Automerge→VFS→iframe path but its 75s-timeout flakiness was the source of every remaining hard failure.

- **Nightly smoke-all + clearer run naming**: the `hub-client-e2e` workflow now runs nightly at 2am EST on `main`
(`run-smoke-all` fires automatically on the schedule trigger). Workflow runs are labelled by mode in the Actions UI:
`E2E — interactive` for every push/PR (the 21-test gate), `E2E — interactive + smoke-all (nightly)` for the
scheduled run, and `E2E — recreate baselines` for snapshot regeneration.

## Test plan

- [x] CI green: 0 failures, 2 flaky, 18 passed, **1.5 min test execution**
- [x] `exp/render-queue-only` experiment: 0 hard failures, 17 flaky, 86 passed, 15.8 min (full suite with smoke-all)
- [x] All 3 comment spec tests pass locally
- [x] New smoke-all fixtures pass on native Rust (`cargo nextest`) and WASM (`vitest`) pipelines
- [x] Visual regression baselines regenerated via `recreate-all-snapshots` after merge (Chromium 145→148)


gordonwoodhull:
**Follow-up: the nightly smoke-all from this PR wasn't running — fixed in `93ff165d` on `main`.**

The first scheduled nightly failed with `Error: No tests found`. Two gaps between the summary and what shipped:

1. **Smoke-all matched zero tests.** The step ran `npx playwright test e2e/smoke-all.spec.ts`, but the base config's `testIgnore` excludes that spec. A positional path is a filter applied after `testIgnore`, so it can't un-ignore the file. Fixed with a dedicated `playwright.smoke-all.config.ts` (mirroring `playwright.visual.config.ts`) that selects the spec via `testMatch`.

2. **Run naming wasn't implemented.** The `E2E — interactive` / `…(nightly)` labels described in the summary never landed as a `run-name:` key. Added it.

The additive structure (21 custom + smoke-all) was already correct, so no change needed there.

One related find: smoke-all is flaky at 2 workers (a manual run came back 5 failed / 25 flaky / 48 passed), but passes cleanly at `workers: 1` — 0 failed / 14 flaky / 64 passed, verified via `workflow_dispatch -f run-smoke-all=true` ([run 27427588049](https://github.com/quarto-dev/q2/actions/runs/27427588049)).

Issue #250, opened 2026-06-02
Wire most workspace TS test suites into CI


gordonwoodhull:
### Context
CI currently runs the TypeScript tests for `hub-client` (via `ts-test-suite.yml` → `npm run test:ci`). The other packages in the npm workspace are built in CI but their own test suites aren't run yet. This looks like a leftover from when TS testing first moved into its own workflow (`bf7e8f42`, "attempt to move ts testing into own GHA workflow") — hub-client got wired up and the rest hasn't been added since.

Most of these suites are healthy today, so this is mostly a matter of connecting them up.

### Suites not currently run in CI (checked locally)

Green as-is:
- `preview-renderer` (208), `preview-runtime` (65), `quarto-automerge-schema` (52), `q2-preview-spa` (8), `quarto-sync-client`, `trace-viewer` (10), `wasm-js-bridge` (10), `q2-demos/kanban` (35)

Needs a small touch before gating in:
- `quarto-hub-mcp` — passes except for a couple of deps not installed in my local checkout; likely green under a clean CI `npm ci`
- `sync-test-harness` — one test sensitive to the same sync/peer-connection flakiness we see in e2e
- `annotated-qmd` — 2 of 156 fail; tracked separately (bd-1d6io)

### A couple of related notes
- `*.integration.test.ts` files are excluded from several packages' default `vitest run`, so they don't run anywhere yet (e.g. most of `q2-preview-spa`'s files).
- Rust doctests aren't run either — nextest doesn't execute them and there's no `cargo test --doc` step. Worth a separate decision.

### Suggested direction
Run the workspace suites in CI (e.g. `npm test --workspaces --if-present`), starting with the green ones so CI stays green, and fold in the others as they're ready. The integration-test and doctest items can be handled separately.

PR #283, opened 2026-06-13
feat(hub-mcp): add wait_for_change long-poll tool

gordonwoodhull:
## What

Adds a `wait_for_change` tool to the Quarto Hub MCP server: a **long-poll** that blocks until a watched file is edited by any collaborator, then returns its new content. This lets an agent react to a live collaborator without busy-polling `read_file`.

```jsonc
wait_for_change({
  project: "<index-doc-id>",
  path: "report.qmd",
  timeout_seconds: 25,    // optional, clamped to 1–55
  since_hash: "sha256:…"  // optional — gap-close, see below
})
```

- Returns `{ changed: true, content, hash }` on an edit, or `{ changed: false, hash }` on timeout (caller just re-calls to keep watching).
- **`since_hash`** closes the gap between polls: if the file already differs from the caller's last-known hash, the call returns immediately — so an edit landing *between* two calls is never missed.
- Read-only tool; available even under `--read-only`.

## Why

Built for agent-in-the-loop collaborative-editing testing — having an agent edit a Hub doc as a second author while a human edits the same file. Polling `read_file` either busy-loops or, if throttled to the disk-flush cycle, lags ~30s. `wait_for_change` reacts in ~1s off the live Automerge sync stream the connection already receives.

## Implementation

`ConnectionManager.waitForChange()` registers a one-shot waiter against the existing `onFileChanged` / `onFileRemoved` sync callbacks and resolves with the new payload (or times out). No new sync machinery — it taps the live `files` map the manager already maintains.

## Tests

- 4 new `ConnectionManager` unit tests: resolve-on-change, timeout, gap-close via `since_hash`, and other-path isolation.
- Protocol tests updated to assert the tool is registered (read-write + read-only lists) — these spawn the real built `dist/index.js`.
- Full suite: **163 passing, 2 skipped**.
- **Verified live end-to-end** against a running hub: watch → collaborator edits in the browser → agent appends to the doc → re-arm, across several rounds, no misses.

## Docs

README gains a **Tools** section documenting every tool, with a subsection on `wait_for_change` semantics (long-poll + `since_hash`).

## Notes / follow-ups (not in this PR)

- Wiring this package's vitest suite into CI is tracked by #250 ("Wire most workspace TS test suites into CI"), which already lists `quarto-hub-mcp` among the suites to fold in. The network-dependent `live:` tests here (they hit `sync.automerge.org`) will need gating when that happens.
- A live two-client integration test (mirroring `sync-test-harness/concurrent-editing.test.ts`) would complement the unit tests; deferred.

### DIFF
--- a/ts-packages/quarto-hub-mcp/README.md
+++ b/ts-packages/quarto-hub-mcp/README.md
@@ -16,6 +16,61 @@ The design lives in
 operators registering a hub for end-user use should consult
 [`claude-notes/instructions/hub-mcp-operator-runbook.md`](../../claude-notes/instructions/hub-mcp-operator-runbook.md).
 
+## Tools
+
+Once configured, the agent has the tools below. Each operates on a project
+identified by its Automerge **index document ID**, passed as the `project`
+argument.
+
+### Reading
+
+| Tool | What it does |
+|------|--------------|
+| `connect_project` | Connect to a project by its index doc ID; returns the file list. Triggers the sign-in flow if the hub requires auth. |
+| `list_files` | List the files in a connected project. |
+| `read_file` | Read a text file's content. |
+| `wait_for_change` | **Long-poll**: block until a file is edited by any collaborator, then return its new content (see below). |
+
+### Writing
+
+| Tool | What it does |
+|------|--------------|
+| `write_file` | Replace a text file's entire content (creates it if absent). |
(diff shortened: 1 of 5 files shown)


cscheid:
This is great! Going into 0.1.1

PR #288, opened 2026-06-15
fix(hub-mcp): quote OAuth URL when opening browser on Windows

cderv:
On Windows, `q2 mcp`'s `authenticate` tool opened the browser to a truncated Google authorization URL — everything after the first `&` was dropped — so sign-in failed with "Erreur 400 : invalid_request — Missing required parameter: redirect_uri" before the user could authenticate.

## Root Cause

The browser launcher built `cmd.exe /c start "" <url>` and spawned it with `windowsVerbatimArguments: false`. Node only wraps an argv element in double quotes when it contains a space, tab, or `"`; an OAuth URL has none of those, so the URL reached `cmd.exe` bare. `cmd.exe` then treats `&` as a statement separator and split the command at the first one, dropping `redirect_uri` and every later parameter before `start` ever saw them.

## Fix

Quote the URL and spawn with `windowsVerbatimArguments: true` so the `&` separators stay literal. Double quotes do not neutralise cmd's `%VAR%` expansion, but the percent-encoded URL is safe because no real environment variable is named like a two-hex-digit token, and the dynamic base64url fields (`state`, `code_challenge`) contain no `%`.

## Test Plan

- [x] `npm test -- src/auth/browser.test.ts` in `ts-packages/quarto-hub-mcp` passes, including the Windows-only cmd.exe round-trip regression test
- [x] On Windows, the `authenticate` browser opens the full URL (`redirect_uri`/`client_id`/`scope` present) instead of truncating at the first `&`

### DIFF
--- a/ts-packages/quarto-hub-mcp/src/auth/browser.test.ts
+++ b/ts-packages/quarto-hub-mcp/src/auth/browser.test.ts
@@ -3,6 +3,7 @@
  * one) and abort-kills-subprocess behaviour.
  */
 
+import { spawnSync } from 'node:child_process';
 import { EventEmitter } from 'node:events';
 import { describe, it, expect, vi } from 'vitest';
 
@@ -16,6 +17,7 @@ describe('browserOpenSpec', () => {
     expect(browserOpenSpec('darwin', URL_WITH_AMP)).toEqual({
       command: 'open',
       args: [URL_WITH_AMP],
+      windowsVerbatimArguments: false,
     });
   });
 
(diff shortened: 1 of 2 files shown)


cscheid:
The robot did all of this; apologies I wasn't able to catch the problems. Do you think there is anything we should be changing in our robot instructions to avoid the class of problems in the future?

cderv:
I was thinking about that. But adding rules means more context in all session. Maybe once there is enough code path robot will know better. 

I was thinking maybe a review skills with common known pattern to use for review to verify if some cross os support and known gotchas are detected. 

For now no problem. I am also learning those Rust problem when encountering them.

PR #289, opened 2026-06-15
fix(xtask): run npm through cmd /C so it resolves on Windows

cderv:
On Windows, every npm-spawning xtask subcommand fails with "program not found": `build-hub-mcp-bundle`, `build-q2-preview-spa`, `build-trace-viewer`, `build-all`, and `verify`. npm ships as `npm.cmd`, and Rust's `std::process::Command` does not search `PATHEXT` or resolve an extensionless name to its `.cmd` shim — only `.exe` may omit its extension. The Mac/Linux CI never exercised this path, so it went unnoticed.

## Fix

Trampoline npm through `cmd /C npm`, letting Windows resolve the real shim via `PATHEXT`. This matches what Tauri (`cross_command`) and wasm-pack (`new_command`) settled on, both having explicitly rejected appending `.cmd`. The logic is centralized in `util::nested_command` behind a small shim allowlist, so `.exe` programs such as cargo are still invoked directly and every call site keeps passing the bare name.

The npm args here are trusted literals, so the cmd/.bat argument-escaping hazard (CVE-2024-24576) does not apply.

## Test Plan

- [x] On Windows, `cargo xtask build-hub-mcp-bundle` runs npm (previously "program not found")
- [ ] On macOS/Linux, npm-spawning xtask commands still work and cargo is unaffected
- [x] `cargo nextest run -p xtask` passes

### DIFF
--- a/crates/xtask/src/build_all.rs
+++ b/crates/xtask/src/build_all.rs
@@ -18,7 +18,6 @@
 
 use anyhow::{Context, Result, bail};
 use std::path::Path;
-use std::process::Command;
 
 /// Configuration for the build-all command.
 #[derive(Default)]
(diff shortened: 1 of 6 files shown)


cderv:
The Hub-Client E2E failure is pre-existing and unrelated to this change. The same test ("editing in Preview view before Monaco mounts updates the Automerge document") timed out on `main` at 09:16 UTC today, before this PR even triggered CI at 09:26 UTC. Our change only touches the Windows xtask path (running npm via `cmd /C`), which has no bearing on this Ubuntu Playwright test.

PR #294, opened 2026-06-16
Error on duplicate keys

hadley:
I can't see a reason that you'd want to allow these in maps, but let me know if you think this needs an extension to the schema to optionally allow.


cscheid:
I agree, and it's weirder than that even: [YAML 1.2 requires key uniqueness (see 3.2.1.1)](https://yaml.org/spec/1.2.2/). I guess it's easier to parse repeated keys and then refuse them than to disallow it, and that's why we need a "validation" error here. I would have expected this to have failed at the parsing stage. I'm completely fine with a hard error message.

PR #295, opened 2026-06-16
Support yaml's FP sentinels

hadley:


cscheid:
This is great, thanks.

To be clear, I expect our handling of `Yaml::Real` to need a good, hard audit. It's not a type that Pandoc uses extensively and I bet that Quarto 2 handles it poorly as well. (Not an excuse for a YAML library, but that's where the path dependence took us).

PR #296, opened 2026-06-16
feat(editor): tree-sitter syntax highlighting for .qmd in Monaco

### LINKED ISSUE #10: bug: Link/image color formatting
I'd expect the whole link to be blue, not dark blue then white then light blue

Image

Related, I'd like for markdown links to be Blue and not brown.

Image

Similarly for image locations

Image

... really I want the link chars to be a different color (light blue, as it _links_ to somewhere else) than the markdown syntax (grey?). I'd love to have the image parameters (`height=400px`) maybe be formatted as if they are code (mid-blue).

To go even further... should the image markdown syntax be a dark brown or a green,  where as the link syntax stays the brown color. (The little ole' `!` at beginning has a very different interpretation to the final product!) It'd help give some visual representation that there are a bunch of links vs a bunch of images.

shikokuchuo:
Fixes #10.

`.qmd` files were mapped to Monaco's built-in `markdown` language, which mis-colours links, ignores YAML frontmatter, and treats Quarto's `{r}`/`{python}` cells and Pandoc constructs (attributes, shortcodes, math) as plain text.

This drives `.qmd` editor highlighting from tree-sitter instead.

The editor requests semantic tokens from a Rust→WASM module. Rust parses the buffer with the `tree-sitter-qmd` grammar and returns `(line, char, length, type)` tokens for three regions — Markdown structure, YAML frontmatter, and code-cell interiors (R/Python/… via the same grammars the renderer uses) — which Monaco colours.

A synchronous Monarch base paints instantly while the async tokens arrive, and fills anything the tokens don't cover.

Key files:

- `crates/quarto-lsp-core/src/tokens.rs` — extractor (parse tree → per-region, per-line UTF-16 tokens); `types.rs` holds the token legend + capture-name → colour-name translator.
- `crates/quarto-highlight/src/captures.rs` — resolver (`Query.captures()` + innermost-wins `flatten_spans`) now shared by the editor and the HTML renderer, so code cells colour identically in both. Also fixes bd-98k6; see the 4 regenerated `*.snap` goldens (nested → flat).
- `crates/wasm-quarto-hub-client/src/lib.rs` — `lsp_get_semantic_tokens` / `lsp_get_token_legend` exports, modelled on the existing `lsp_get_symbols`.
- `hub-client/src/services/monacoProviders.ts`, `components/quartoTheme.ts` — Monaco provider, `quarto-{light,dark}` themes, Monarch base.

Code-cell colours mirror the rendered `hl-*` CSS (a test locks the two tables together); every theme rule carries a `qmd.` prefix so it can't recolour other languages in the editor.

Design + phase notes: `claude-notes/plans/2026-06-10-monaco-tree-sitter-highlighting.md`.


shikokuchuo:
Before / after captures:

compare-structure-light
compare-code-cells-light

cscheid:
I particularly like the calls to the R and Python tree-sitter grammars for their own highlighting!

cscheid:
I don't think we should highlight comments like that, though; they should be, well, comments!

shikokuchuo:
> I don't think we should highlight comments like that, though; they should be, well, comments!

I assume you mean comments within the qmd - those are now fixed in [b2dbd3c](https://github.com/quarto-dev/q2/pull/296/commits/b2dbd3ce5561dd382539d6bce1098a56670b877c) and appear grey+italic. We weren't emitting any `qmd.markup.comment` tokens in the first place.

shikokuchuo:
There are some follow-ups, but they are not blocking for this PR: 1. {.python} chunks, 2. yaml options in code chunks etc.

EDIT: Follow-up 1 is implemented in #309 and 2 in #311.

PR #298, opened 2026-06-16
Add a compact output mode (for LLMs)

hadley:


cscheid:
... huh, this is neat. Our JSON mode --- which q2 shares more generally --- is meant for machine consumption, but I do see how this could help LLMs. Is this urgent to merge? I think I might want to first send an agent to investigate if we could adopt this more broadly in q2 (which would be a pretty big change, but if I'm not the one editing the files...)

hadley:
No rush! We'll use it upstream whenever it's ready.

cscheid:
Eh, I'll just merge it here and we'll smooth over wrinkles when it comes up.

PR #302, opened 2026-06-17
chore(xtask): rename strand branch prefix beads/ -> braid/

shikokuchuo:
bd-yjh1y117.

The local topic-branch prefix for braid strands was still `beads/` — a historical git namespace left over from before the beads→braid migration.

The name is misleading: nothing about beads is involved, and the strand lives in the braid skein. The prefix is just a git label that mirrors the strand id.

This renames the emitted prefix to `braid/` via a shared `strand_branch()` helper in `create_worktree.rs`, used by both `create-worktree` and `switch-task`. The worktree conventions and the affected skill docs are updated to match.

Out of scope, deliberately left alone: remote work-type prefixes (`bugfix/`, `feature/`) are unchanged, and `.beads/`-directory references (the old beads-tool data dir) stay as historical context.

### DIFF
--- a/.claude/rules/worktrees.md
+++ b/.claude/rules/worktrees.md
@@ -36,7 +36,7 @@ the *shape* of the work, not on the strand boundary.
 
   That switches the current worktree to `feature/q2-preview-command`,
   fast-forward-pulls (so any sibling sub-tasks that merged in the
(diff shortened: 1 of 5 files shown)


cscheid:
Yes, this is a good idea. I think we should consider going further. I often end up with branch names that are unreadably long, and not very useful as a name.

PR #306, opened 2026-06-17
chore(deps): remove unused dependency declarations

shikokuchuo:
Removes **59 unused dependency declarations across 27 crates** — manifest hygiene, no source changes.

## What we used
[`cargo-machete`](https://github.com/bnjbvr/cargo-machete) (`--with-metadata`) to find unused deps, then Claude verified every hit "by hand" before removing it.

## Intentionally risk-averse
We removed only declarations with zero usage and left anything that *looks* unused but isn't. Kept on purpose:

- `openssl-sys` (`quarto`) — link-only, behind the `vendored-openssl` feature
- `quarto-error-message-macros` (`quarto-parse-errors`) — proc-macro; derive name ≠ crate name
- `tokio` (`quarto-lsp-core`) — macro-based usage
- `sha2` (`quarto-sass`) — used in `build.rs` (machete doesn't scan build scripts)
- `cc` + the wasm/fuzz/patch-crate deps — left untouched

## Impact
Only `pretty_assertions` (+ `diff`) drops from the build graph entirely; the rest were redundant declarations of crates still used elsewhere.

## How we tested
Full `cargo xtask verify` — all 14 steps green, covering `cargo build --workspace`, `cargo nextest run --workspace`, the WASM rebuild, and the hub-client build/tests. The all-targets build is itself the proof: any test relying on a removed dev-dep would have failed to compile.

Strand: bd-e3lv7eg3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -658,7 +658,6 @@ version = "0.3.0"
 dependencies = [
  "comrak",
  "hashlink",
- "insta",
  "pampa",
  "proptest",
  "quarto-pandoc-types",
@@ -1133,12 +1132,6 @@ dependencies = [
  "rustc_version",
  "syn",
 ]
-
-[[package]]
-name = "diff"
-version = "0.1.13"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
 
 [[package]]
 name = "digest"
@@ -2936,7 +2929,6 @@ dependencies = [
 name = "pampa"
 version = "0.0.0"
 dependencies = [
- "ariadne",
  "base64 0.22.1",
  "clap",
  "comrak",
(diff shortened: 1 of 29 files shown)


cscheid:
I think I'm missing something here, because a lot of the lines being removed reference crates we do need (`ariadne`, for example). Can you clarify what this is trying to do?

shikokuchuo:
I had Claude just double-check on Ariadne as you mentioned it.

- `pampa` declared `ariadne = "0.6"` but has no use for `ariadne` anywhere in its source. The only real consumer is `quarto-error-reporting`;
- `pampa` gets `ariadne` formatting transitively through it. So pampa's declaration was redundant — `ariadne` stays in `Cargo.lock` and `pampa` still builds.

I was just trying to see how much we could clean up 'for free' - and it seems not much: the only 2 crates **removed** are `diff` and `pretty_assertions` as nothing used them at all.

cscheid:
Ok, got it. I'm fine with merging this, I just wanted to know the context (we also have a lockfile conflict somehow).

shikokuchuo:
I just wanted to share the result really as I'd attempted it... I've re-based so should be ok to merge.

Also fine not to merge, and leave this all for later...

PR #311, opened 2026-06-18
feat(lsp): highlight cell-option #| lines as YAML in editor

shikokuchuo:
`#|` cell options now highlight as YAML in the editor instead of as a flat comment — the key reads as a property, the value as a string/boolean, and the `#|` marker stays a comment. Only the leading run is treated as options (matching Quarto's top-of-cell rule); a `#|` after code stays a plain comment.

```r
#| echo: true        # before: one grey comment  →  after: "echo" property, "true" boolean
library(ggplot2)     # full r highlighting, unchanged
```

### DIFF
--- a/crates/quarto-lsp-core/src/tokens.rs
+++ b/crates/quarto-lsp-core/src/tokens.rs
@@ -287,7 +287,74 @@ fn code_cell_tokens(code_block: Node<'_>, content: &str) -> Vec<ByteToken> {
 
     let body_start = interior.start_byte();
     let body = node_text(interior, content);
-    embedded_body_tokens(&lang, body, body_start)
+
+    // A leading run of `#|` lines is a Quarto cell-option block — YAML, not
+    // code. Highlight it as yaml; hand only the remaining code to the cell grammar.
+    let mut out = Vec::new();
+    let directive_end = leading_directive_byte_len(body);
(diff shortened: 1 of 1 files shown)


cscheid:
This is a good improvement, but let me add a couple of fun wrinkles:

````
```{r}
#| foo:
#|   bar:
#|     baz: >
#|       A _YAML_ string.
#|   array:
#|     - 1
#|     - 2
#|   also_array: [
#|     1, 2, 3
#|   ]
#| echo: true
cat("hello, world")
```
````

In other words, you can't parse each line as YAML separately.

shikokuchuo:
Actually it works quite well as we don't need it to parse correctly - the colour is determined by the single-line lexical structure so it looks mostly right.

Only `also_array` in the above example is off. Something we can live with?

cscheid:
> Actually it works quite well as we don't need it to parse correctly - the colour is determined by the single-line lexical structure so it looks mostly right.

The thing is that "mostly right" is not going to be enough once we start showing yaml validation errors for those code blocks (which Quarto 2 _can_ do).

shikokuchuo:
I was going to actually change it to parse as a block - which doesn't really cost us anything as it all happens within a single WASM call. I'll add it as a follow-up!

PR #312, opened 2026-06-18
fix(lsp): parse #| cell-option block as one YAML document

shikokuchuo:
Follow-up to #311 (the editor-only `#|` cell-option YAML highlighting), which was merged before this commit landed.

Parses the whole leading `#|` option block as a **single** YAML document instead of line-by-line: the block is reassembled (markers stripped, indentation preserved), highlighted once, and the spans are mapped back per line. 

Wrt. https://github.com/quarto-dev/q2/pull/311#issuecomment-4742278672

This makes genuinely multi-line YAML highlight correctly — block scalars (`>`) and flow collections split across lines (`[ … ]`), where the per-line approach mis-typed tokens (e.g. `also_array` now reads as a property, `1, 2, 3` as three numbers, the closing `]` as punctuation).

Editor-only; two regression tests added.

### DIFF
--- a/crates/quarto-lsp-core/src/tokens.rs
+++ b/crates/quarto-lsp-core/src/tokens.rs
@@ -320,14 +320,32 @@ fn leading_directive_byte_len(body: &str) -> usize {
     end
 }
 
+/// Where one `#|` line's YAML content lives in both coordinate systems: its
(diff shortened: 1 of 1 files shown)


cscheid:
I should keep sending progressively weirder problems your way!

PR #329, opened 2026-06-23
fix(doctemplate): preserve line-ending convention in multiline directives

### LINKED ISSUE #157: doctemplate: CRLF templates produce extra blank lines around multiline $if$ / $for$
On Windows with `core.autocrlf=true`, doctemplate templates with CRLF line endings render with extra blank lines around every multiline `$if(...)$` and `$for(...)$` directive. The symptom currently shows up only as 8 failing tests in `quarto-doctemplate`, but the underlying engine bug means a Windows author of any Quarto template gets visibly wrong output in every format the doctemplate engine drives. We don't have Windows CI yet, so this hasn't surfaced for the rest of the team in `cargo nextest run`.

I reproduced this on Windows at `ebd04493`.

## Reproducer

```rust
let crlf_source = "before\r\n$if(show)$\r\ncontent\r\n$endif$\r\nafter\r\n";
let template = Template::compile(crlf_source).unwrap();
let mut ctx = TemplateContext::new();
ctx.insert("show", TemplateValue::Bool(true));
let result = template.render(&ctx).unwrap();
// expected: "before\ncontent\nafter\n"
// actual:   "before\r\n\r\ncontent\r\n\r\nafter\r\n"
```

Same shape on `$for$` / `$endfor$`, nested `$if$` / `$for$`, and `$else$`.

## Root cause

`normalize_multiline_directives` (run after tree-sitter parsing) detects "directive on its own line" by checking the first character of the body Literal. The detection helpers hardcode `'\n'`:

https://github.com/quarto-dev/q2/blob/ebd044938a18e2984a9e5439f88ef5a3aa162326/crates/quarto-doctemplate/src/parser.rs#L1067-L1095

For CRLF input the body Literal starts with `\r\n`, so `starts_with('\n')` returns false, `is_multiline` stays false, and the branch that consumes the leading and trailing newlines around the directive never runs:

https://github.com/quarto-dev/q2/blob/ebd044938a18e2984a9e5439f88ef5a3aa162326/crates/quarto-doctemplate/src/parser.rs#L973-L1056

Pandoc's `doctemplates`, by contrast, is parser-aware. Its endline parser accepts all three conventions and returns whatever was matched, so multiline directive consumption is line-ending-agnostic and the output preserves the input convention:

```haskell
pLineEnding = P.string "\n" <|> P.try (P.string "\r\n") <|> P.string "\r"
isSpacy '\r' = True
pLit = P.many1 (P.satisfy (\c -> c /= '$' && c /= '\n' && c /= '\r'))
```

https://github.com/jgm/doctemplates/blob/master/src/Text/DocTemplates/Parser.hs#L262-L263

Pandoc's `--eol=crlf|lf|native` is a separate writer option layered on top.

## Constraints I see for q2 on Windows

CRLF input must render correctly. Output should preserve the input line-ending convention — silently rewriting bytes mid-pipeline diverges from Pandoc and surprises Windows users whose rest-of-file convention is CRLF. Source spans (`node.start_byte()`, `start_position`) need to keep mapping back to the on-disk file or diagnostics drift.

A one-line ingress normalize (CRLF→LF before parsing) is out: it loses the input convention and shifts every byte offset by one per preceding CRLF.

## Approaches

We could teach the Rust normalization helpers to recognize `\r\n` and `\r` in addition to `\n`, plus an audit of `tree-sitter-doctemplate` to see whether the grammar needs the same alternation or whether the Rust pass alone is enough. Bytes preserved end-to-end. Same shape of work as #139 did for `tree-sitter-qmd` pipe tables, smaller scope.

We could also normalize CRLF→LF for the parser internally with a side-table mapping normalized→original byte positions for diagnostics, then re-emit the input convention on render. More machinery, easier to forget the side-table when adding new diagnostics.

## Open question

Is "preserve input line-ending convention end-to-end" the policy we want for q2 on Windows? Or would we rather always normalize to LF on output, or expose a writer-side option like Pandoc's `--eol`?

This is broader than `quarto-doctemplate` — pampa output, the JSON / native writers, and any future tree-sitter grammar will face the same question. Picking a policy here sets the precedent.

If the answer is "preserve input convention", I'll scope the `tree-sitter-doctemplate` audit, add a CRLF regression test that builds the input in-process so Linux CI catches future regressions (same pattern as `pipe_table_crlf_matches_lf` from #139), and update the 8 affected `quarto-doctemplate` tests. Internal tracker is bd-1d3e.

cderv:
On Windows with `core.autocrlf=true`, doctemplate templates with CRLF line
endings render an extra blank line around every multiline `$if$` / `$for$` /
`$else$` directive. It surfaced as 8 failing tests in `quarto-doctemplate`, but
the engine bug means any Windows author of a Quarto template gets visibly wrong
output in every format the doctemplate engine drives. There is no Windows CI
yet, so it never showed up in `cargo nextest run` for the rest of the team.

## Root Cause

`normalize_multiline_directives` reproduces Pandoc's behavior of swallowing the
newlines around a directive that sits on its own line. Its detection and
stripping helpers hardcoded `'\n'` (`first_node_is_newline_literal`,
`strip_leading_newline_from_node`). With CRLF input the directive's body literal
starts with `\r\n`, so `starts_with('\n')` returns false, the directive is never
treated as multiline, and the newline-consumption branch never runs — leaving
the doubled blank line.

The bug is in the engine and reproduces with in-memory `Template::compile`; it
is not a file-read concern. An earlier attempt that normalized CRLF→LF at file
ingestion was discarded: it papered over this engine bug and is the approach
#157 explicitly rules out.

## Decision: preserve the input convention

CRLF input now renders CRLF output, LF renders LF — the input line-ending
convention is preserved end-to-end, matching Pandoc's parser-aware
`doctemplates`, whose endline parser accepts all three conventions.

Two alternatives were rejected:

- Ingress-normalize CRLF→LF before parsing — silently rewrites bytes and shifts
  every source offset by one per preceding `\r`, which drifts diagnostics.
- Always emit LF — diverges from the rest of a Windows author's file, and still
  needs the engine fix anyway since the bug fires in-memory.

This sets the precedent for the other writers (pampa output, JSON/native) and
future grammars: conventions are preserved, not normalized. A writer-side
`--eol` override, like Pandoc's, stays a separate future concern.

## Fix

Detection and stripping now recognize `\r\n`, `\r`, and `\n` each as one
line-ending unit (new `leading_line_ending_len`), so the full sequence is
consumed and the surrounding CRLF content is left intact. Source bytes and
offsets are untouched.

The `.template` test fixtures are pinned to LF via `.gitattributes` so the
LF-asserting fixture tests stay deterministic regardless of `core.autocrlf`; the
CRLF path is covered by new in-process tests that build the input in source, so
Linux CI catches future regressions (same pattern as #139 for tree-sitter-qmd
pipe tables).

The `tree-sitter-doctemplate` grammar needs no change: its `text` token
(`/[^$]+/`) already captures CRLF verbatim, so `\r\n` reaches the Rust
normalization pass intact.

## Test Plan

- [ ] `cargo nextest run -p quarto-doctemplate` — 200/200, including new `crlf_preservation` tests
- [ ] CRLF template renders CRLF output with no doubled blank lines
- [ ] LF template output unchanged
- [ ] `cargo xtask verify --skip-hub-build`

Fixes #157

### DIFF
--- a/claude-notes/plans/2026-06-23-doctemplate-crlf-line-ending-preservation.md
+++ b/claude-notes/plans/2026-06-23-doctemplate-crlf-line-ending-preservation.md
@@ -1,0 +1,75 @@
+# doctemplate CRLF line-ending preservation (bd-1d3e / #157)
+
+## Policy decision
+
+**Preserve the input line-ending convention end-to-end.** CRLF input renders
+CRLF output; LF input renders LF output. No ingress normalization, no silent
+byte rewriting. (Chris, 2026-06-23. Matches Pandoc's parser-aware doctemplates;
+`--eol` writer override is a separate, future concern.)
+
+This sets precedent for pampa output, the JSON/native writers, and future
+tree-sitter grammars: line-ending conventions are preserved, not normalized.
+
+## Root cause
+
+`normalize_multiline_directives` (`crates/quarto-doctemplate/src/parser.rs`)
+detects a "directive on its own line" and swallows the surrounding newlines, to
+match Pandoc. Its three helper functions hardcode `'\n'`:
+
+- `first_node_is_newline_literal` / `is_first_child_newline_literal` — detection
+  via `lit.text.starts_with('\n')` (parser.rs:1093-1098).
+- `strip_leading_newline_from_node` — `starts_with('\n')` then `text[1..]`
+  strips a single byte (parser.rs:1115-1120).
+
+For CRLF input the body Literal starts with `\r\n`, so detection returns false,
+`is_multiline` stays false, and the newline-consumption branch never runs →
+doubled blank lines around every multiline `$if$` / `$for$` / `$else$`.
+
+The bug is in the engine and reproduces with in-memory `Template::compile` —
+**not** a file-read concern. (An earlier ingress-normalize attempt was reverted:
+it was the approach #157 explicitly rules out, and it masked this engine bug.)
+
+## Fix
+
+Make the three helpers treat `\r\n`, `\r`, and `\n` each as one leading
+line-ending unit:
+
+- detection: body's first Literal starts with any of `\n` / `\r\n` / `\r`.
+- strip: remove the full leading sequence (2 bytes for `\r\n`, else 1).
+
+Single chokepoint — all detection/stripping routes through these helpers — so
+both `$if$` and `$for$` (and nested/else) are fixed at once. CRLF bytes in the
+surrounding literals are preserved, so output keeps the input convention.
+
+## tree-sitter-doctemplate audit
+
+The buggy output already contained `\r\n` inside the body Literal, so the
+grammar tokenizes CRLF into literal text fine; the Rust pass alone should
+suffice. Confirm no grammar change is needed (no `\r`-specific tokenization gap).
+
+## Test fixtures
+
+`tests/pandoc-equiv/*.template` and `test-fixtures/*.template` are checked out
+CRLF on Windows (`core.autocrlf=true`). Their tests assert LF output. Pin these
+fixtures to LF via a scoped `.gitattributes` (`*.template eol=lf`) + renormalize,
+so they are LF on every platform and the existing LF assertions verify
+LF-preservation. The CRLF path is covered by new in-process tests.
+
+## Work items
+
+- [x] TDD: in-process CRLF regression tests — `Template::compile(crlf_source)`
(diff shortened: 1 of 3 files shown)


cderv:
We discussed this I think last time, and hopefully I went in the right direction to support all EOL from the input

cscheid:
Thanks, this looks good!

PR #332, opened 2026-06-23
fix(pampa): handle Windows paths in synthetic Lua io/dofile resolution

cderv:
On Windows, 9 `io_wasm` unit tests in `pampa` fail. The synthetic Lua `io` table is exercised natively with real temp files, and Windows paths break it two ways.

## Root Cause

Two distinct problems in the test path:

1. The test scripts interpolate `file_path.display()` into Lua source, so a Windows path like `C:\Users\…` injects backslash escape sequences (`\U`, `\t`) and Lua fails with a SyntaxError.
2. `io.open` classified absolute-vs-relative with `path.starts_with('/')`. A Windows absolute path (`C:/…`) isn't matched, so it gets the VFS `/project/` prefix and the read fails.

`dofile`/`loadfile` resolution had a related but separate defect: it used `Path::is_absolute()`, which returns `false` on `wasm32-unknown-unknown` for rooted `/foo` paths — std treats the target as neither unix nor wasi, and it carries no prefix. A rooted VFS path was therefore mis-resolved, a real bug on the only target that runs that code.

## Fix

Add `quarto_util::is_rooted` (a `Path::has_root()` wrapper, correct on native and WASM) and use it at both VFS-resolution sites, replacing the two divergent checks with one predicate. Embed forward-slash paths in the affected tests. WASM behavior is unchanged: in WASM these sites only ever see `/`-rooted or relative paths, for which `has_root()` and the old checks agree.

`quarto-util` becomes a normal (not dev-only) dependency of `pampa`; it was already compiled for wasm32 via `quarto-core`, so no new build surface.

## Test Plan

- [ ] `cargo nextest run -p pampa` green on Windows (9 io_wasm tests + 2 new dofile resolver tests)
- [ ] `cargo nextest run -p quarto-util` green (is_rooted, incl. Windows drive-path)
- [ ] Full `cargo xtask verify` green

### DIFF
--- a/claude-notes/plans/2026-06-23-bd-vxl8-windows-lua-path.md
+++ b/claude-notes/plans/2026-06-23-bd-vxl8-windows-lua-path.md
@@ -1,0 +1,338 @@
+# bd-vxl8: Windows path handling in pampa synthetic Lua io/dofile — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the 9 failing `io_wasm` tests pass on Windows and fix a latent wasm-only `dofile` path-resolution bug, by replacing two inconsistent inline absolute-path checks with one shared, cross-target-correct predicate.
+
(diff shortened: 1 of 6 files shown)


cscheid:
> The test scripts interpolate file_path.display() into Lua source, so a Windows path like C:\Users\… injects backslash escape sequences (\U, \t) and Lua fails with a SyntaxError.

🤦🤦🤦🤦🤦

cscheid:
I think this looks good!

PR #347, opened 2026-06-26
Gate native-pampa round-trip tests behind opt-in env var (bd-d8nol0xn)

cscheid:
## Summary

The two bd-sjb4pzx8 round-trip tests shell out to the native `pampa` binary (build + per-fixture subprocess over temp files) and are slow + **intermittently flaky** under the parallel load of a full `vitest run` / `cargo xtask verify` — a transient process/IO race around concurrent `cargo build --bin pampa` and temp dirs — while passing in isolation and on re-run. They are measurement/fidelity oracles, not routine gates:

- `tiptap-roundtrip-spike/roundtrip.test.ts` — the throwaway feasibility spike.
- `richtext/roundtrip.test.ts` — the production round-trip fidelity guard (via `test-utils/pampaOracle`).

## Change

Gate both with `describe.skipIf(!process.env.QUARTO_RUN_PAMPA_ROUNDTRIP)` so they're **skipped in routine runs** and **runnable on demand**:

```
QUARTO_RUN_PAMPA_ROUNDTRIP=1 npx vitest run src/q2-preview/richtext/roundtrip.test.ts
```

(The richtext guard keeps its existing `pampaAvailable()` skip too.) Each file's header documents the opt-in.

## Scope / what's kept

Only the two native-pampa **oracle** tests are gated. The fast jsdom rich-text unit tests of shipping behavior — `caretFromClick.test.ts`, `RichTextEditor.caret.integration.test.tsx` — are untouched and still run every time.

## Verification

- Default `vitest run` (preview-renderer): **468 passed | 32 skipped**, ~0.7s, **no pampa subprocesses** → can no longer flake (was 500 with the oracle running, ~6–23s + noisy oracle output).
- Opt-in: `QUARTO_RUN_PAMPA_ROUNDTRIP=1 … richtext/roundtrip.test.ts` → 15 passed (shells out to pampa as before).
- `cargo xtask verify --skip-hub-build` green.

Strand: bd-d8nol0xn (discovered via the flake hit in bd-pvcnea83 / PR #346).

### DIFF
--- a/ts-packages/preview-renderer/src/q2-preview/richtext/roundtrip.test.ts
+++ b/ts-packages/preview-renderer/src/q2-preview/richtext/roundtrip.test.ts
@@ -4,8 +4,12 @@
 //
 //   qmd -> pampa AST -> astToDoc -> docToMarkdown -> pampa AST -> compare
 //
-// Skips when native pampa can't be built (Rust-less CI). Run locally:
-//   cd ts-packages/preview-renderer && npx vitest run src/q2-preview/richtext/
+// DISABLED BY DEFAULT (bd-d8nol0xn): shells out to the native `pampa` binary via
+// the test-utils oracle (slow, and intermittently flaky under the parallel load of
+// a full `vitest run` / `cargo xtask verify`). This is a fidelity oracle, not a
+// routine gate. Opt in (and only runs when native pampa is buildable):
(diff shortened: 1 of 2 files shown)


cscheid:
I don't believe the bot very much here, but these are not especially important tests (famous last words..)

PR #352, opened 2026-06-30
feat(hub-client): Support accessing the hub under a subpath

atheriel:
This commit makes various changes to let the hub be served behind a reverse proxy that mounts it under a subpath (e.g. `/subpath`) rather than at an origin root. This is controlled by a new build-time knob, `VITE_HUB_BASE_PATH`.

Unit tests are included.


cscheid (approved):
Looks great! When this passes CI we'll merge and fire off a new q2 binary release.

PR #358, opened 2026-07-02
Remove unused error_node_analysis integration test

cderv:
The test never asserted anything -- every line was a println! diagnostic -- and it read a personal fixture path (`~/today/categorical-predictors.qmd`) that only existed on its original author's machine. Everywhere else it silently hit the file-not-found skip and did nothing.

It also crashed on Windows shells without a `HOME` env var (PowerShell, cmd), since the `.unwrap()` on `env::var("HOME")` ran before that skip check.

Git history traces it to a bulk "make tests pass" commit that swept in a one-off diagnostic script, later carried into the permanent integration-test suite by the test-consolidation refactor without re-evaluation.


crates/pampa/tests/integration/error_node_analysis.rs
```diff
@@ -1,267 +0,0 @@
-/*
- * error_node_analysis.rs
- * Temporary test to analyze ERROR nodes in tree-sitter parse tree
```
cderv, line 3: It says temporary so I assume this is ok to remove ?

cscheid: Yeah, clearly not something we should have added. Thanks.

cscheid (approved):

PR #359, opened 2026-07-02
Support fnm and nvm-windows Node discovery on Windows

cderv:
When a GUI app like Claude Desktop launches `q2 mcp`, the server starts with a minimal environment and no shell-sourced PATH. fnm on Windows mutates PATH only inside an interactive shell and never persists it, so a GUI-launched host never sees an fnm-managed Node. The layer-3 well-known fallback in `crates/quarto-mcp-launcher/src/node.rs` only probed Program Files and Volta on Windows, so discovery failed outright — `node: NOT FOUND` even with an fnm default installed.

## Fix

Probe the two Windows-native version managers in `default_well_known`:

- fnm under `dirs::config_dir()/fnm` — the `aliases/default` junction and the `node-versions` tree, which has no `bin/` subdir on Windows (unlike unix fnm).
- nvm-windows under `%NVM_HOME%` — a flat `vX.Y.Z/node.exe` layout with no subdir.

Rather than parameterize the version-tree walker, `highest_version_node` now tries a superset of sub-paths covering all four layouts (unix nvm, unix fnm, Windows fnm, nvm-windows). A sub-path for the wrong OS is never a file and is skipped, so unix resolution is unchanged. Adding Windows callers also makes the helper live on Windows, clearing a latent dead_code warning.

nvm-windows persists its active version to PATH via the registry, so most GUI apps already inherit it; that probe is defense-in-depth. The confirmed-broken case is fnm.

## Test Plan

- [ ] `highest_version_node` selects the newest version across all four layouts (unit tests)
- [ ] With PATH stripped to Windows system dirs and an fnm default set, `q2 mcp --launcher-info` reports the fnm Node instead of NOT FOUND
- [ ] Unix discovery is unchanged

### DIFF
--- a/claude-notes/plans/2026-07-02-fnm-nvm-windows-node-discovery.md
+++ b/claude-notes/plans/2026-07-02-fnm-nvm-windows-node-discovery.md
@@ -1,0 +1,287 @@
+# fnm + nvm-windows Node discovery for `q2 mcp` (bd-e1ger90c)
+
+## Overview
+
+`crates/quarto-mcp-launcher/src/node.rs` resolves an ambient Node.js
+for `q2 mcp`. Layer 3 (`default_well_known()`) is a fallback probe of
+known install locations, used only when `QUARTO_NODE` is unset and
+`node` isn't on `PATH`. On Windows this layer only checks Program
+Files and Volta — no fnm, no nvm-windows.
+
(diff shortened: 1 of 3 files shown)


cscheid:
Thanks!

PR #380, opened 2026-07-07
fix(quarto-citeproc): stop false CSL lockfile mismatches

cderv:
When csl_validate_manifest ran after a rebase, it reported "Lockfile mismatch: tests/csl_conformance.lock" even though enabled_tests.txt and test-data/csl-suite/ were unchanged and regenerating the lockfile produced a byte-identical file.

## Root Cause

Two independent bugs, both making the check flaky rather than authoritative:

1. `csl_validate_manifest` was generated by `build.rs`, which baked an "expected" lockfile string into the compiled test binary at build-script run time and compared it against the live file at test-run time. If Cargo didn't rerun `build.rs` when it should have, the baked value went stale relative to disk - a time-of-check/time-of-use gap.
2. Separately, on Windows checkouts with the common `core.autocrlf=true` setting, `tests/csl_conformance.lock` gets checked out with CRLF line endings, while the generator that produces the "expected" string always emits LF-only. The raw byte compare then fails on line endings alone, with no real content divergence - `crates/quarto-citeproc` had no `.gitattributes` pinning the file, unlike `crates/pampa`, which already pins its fixtures the same way.

## Fix

`csl_validate_manifest` is now a hand-written runtime test in `tests/integration/csl_conformance.rs` that reads `test-data/csl-suite/` and `enabled_tests.txt` live at test-run time via `env!("CARGO_MANIFEST_DIR")` - a compile-time path constant, not a baked computed value - and compares against the lockfile in the same instant. `build.rs` keeps only per-fixture test codegen. Duplicate/nonexistent-entry detection moves from a soft `cargo:warning` to a hard test failure. `UPDATE_CSL_LOCKFILE=1` auto-update behavior is unchanged.

`crates/quarto-citeproc/.gitattributes` now pins `tests/csl_conformance.lock` and `tests/enabled_tests.txt` to `eol=lf`, matching the existing `crates/pampa/.gitattributes` convention. This overrides `core.autocrlf` for these two paths in both directions, so regenerating the lockfile from Windows stays safe too.

### DIFF
--- a/claude-notes/plans/2026-07-07-csl-lockfile-windows-rerun-if-changed.md
+++ b/claude-notes/plans/2026-07-07-csl-lockfile-windows-rerun-if-changed.md
@@ -1,0 +1,150 @@
+# CSL manifest lockfile false-positive on Windows incremental builds
+
+Strand: bd-2w80 ("Investigate CSL manifest test failure after rebase") — **closed**.
+
+## Resolution (final)
+
+Neither research agent's mechanism theory panned out. Empirical repro (add
+then remove a fixture file, no `cargo clean`) showed directory-level
+`rerun-if-changed` correctly triggers a rebuild both ways on cargo
+1.97.0-nightly — the "directory watch misses changes" theories (both the
+pre-1.50-citation one and the mtime-granularity one) were never confirmed as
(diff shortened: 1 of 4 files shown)


cderv:
Retrying, but I think those CI errors are from main too

PR #397, opened 2026-07-17
hub-client: connection-gated auth + local-first default, with offline editing of cached hub projects

shikokuchuo:
Makes the hub client local-first, with authentication gated on connecting to a hub rather than on entry.

**Local-first default**
- The LoginScreen entry gate is gone; first run auto-creates a local project set.
- Projects without a sync server are created and edited entirely in-browser against a storage-only Repo, authored under a persisted per-browser local actor.

**Connection-gated auth**
- Sign-in is now an opt-in "Connect to a hub" flow.
- A hub project that needs a session we don't have prompts for sign-in instead of silently no-oping.

**Offline editing of cached hub projects**
- A *cached* hub project opens and is editable offline/logged-off under the local actor.
- On reconnect, the sync client re-authors open handles to the server-trusted HMAC actor and bridges identities so the offline+online history reads as one author.

**Also included**
- Restores the editable Sync Server URL field in Create/Import, with follow-up fixes for the local-creation default and a flash-and-bounce teardown regression.
- Local-first E2E spec and `LOCAL-FIRST.md` user docs.
- local-prod static port configurable via `PORT`.

Plans: `claude-notes/plans/2026-07-06-hub-client-connection-gated-local-first.md` and `2026-07-15-hub-client-offline-cached-hub-projects.md` (epic bd-xxjy9yfp), under the auth-reshape umbrella.


shikokuchuo:
I've re-sequenced the auth work, and will be implementing this slightly differently as a result.

PR #423, opened 2026-07-27
Float/layout DOM class taxonomy for HTML figures, tables, listings (bd-hcp8m3ve)

cscheid:
Emits the Q1-verbatim float/layout DOM class taxonomy for HTML figures, tables
and listings, so Q2's output carries the classes Q1's CSS selects on.

**This branch was written 2026-07-21 and sat unmerged.** It was recovered during
the in-progress strand audit (bd-a0eyjshu) — not subsumed by anything on `main`,
and it merges cleanly. Rebased on current `main` (39 commits) and re-verified
before opening.

## Why this matters now

PR #408 ported `_quarto-rules.scss`'s CSS to Q2 — but not the DOM those rules
target. None of `quarto-figure`, `quarto-float`, `quarto-float-caption`, or
`data-qf-*` exist anywhere on `main` today, so a chunk of that CSS currently has
nothing to select. This supplies the DOM.

It also unblocks **bd-9fz5fweg** ("Figures/floats/layout-panels CSS from
`_quarto-rules`"), which carries an explicit `blocks` dependency on bd-hcp8m3ve
under epic **bd-4doe9lvt**.

## What's in it

- **Design contract** — `claude-notes/designs/float-layout-class-taxonomy.md`
  pins the Q1 DOM verbatim from `floatreftarget.lua` / `layout/html.lua`.
- **Figure floats** — outer `div.quarto-float.quarto-figure.quarto-figure-<align>`,
  attrred `<figure>`, synthesized `<figcaption>` + `aria-describedby`.
- **Table/listing floats** — shape 1 (`figure.quarto-float-tbl`), listings get
  `listing` + left align.
- **Standalone captioned figures** — shape 2 wrapper, `fig-align` harvested.
- **React preview parity** — `Figure.tsx` mirrors the pampa writer's figcaption
  synthesis, so preview and render agree.
- Mechanism: transform-side native-node construction in `render_float_ref_target`
  (Finalization phase), figcaption metadata carried as stripped `data-qf-*` kvs.
  Per-repo policy, no DOM postprocessor.

Design decisions signed off 2026-07-21: drop Q1's figcaption uuid in favour of
`<float-id>-caption` with an AST-wide collision check; emit bare `quarto-float`;
five `data-qf-*` kvs; table-float DOM change ships in the same PR (Q2 is 0.x).

## Verification

`cargo xtask verify` — **all 14 steps pass**, exit 0:

- 10511/10511 Rust tests (197 skipped), clippy under `-D warnings`
- WASM + hub-client build, all TS suites, preview-* packages
- **Zero snapshot churn**

End-to-end through the real binary (`cargo run --bin q2 -- render`) on a fixture
with a crossref figure, a standalone `fig-align="right"` figure, and a `#tbl-`
table. Observed output:

```html
<div id="fig-one" class="quarto-float quarto-figure quarto-figure-center">
<figure class="quarto-float quarto-float-fig">
<div aria-describedby="fig-one-caption">
<img src="..." alt="A crossref figure caption" />
</div>
<figcaption id="fig-one-caption" class="quarto-float-caption-bottom quarto-float-caption quarto-float-fig">
A crossref figure caption
</figcaption>
</figure>
</div>
<p>See <a href="#fig-one" class="quarto-xref">Figure 1</a>.</p>
<div class="quarto-figure quarto-figure-right">
<figure>
<img src="..." alt="A standalone caption" />
<figcaption>A standalone caption</figcaption>
</figure>
</div>
```

Crossrefs resolve ("Figure 1", "Table 1"); `fig-align` lands as
`quarto-figure-right`; sections are not float-wrapped.

## ⚠️ Known issues found by that e2e run — please read before merging

Neither is caught by the 10511-test suite; both were found only by rendering
through the binary.

1. **Table floats render their caption twice** — filed as **bd-4m2n6qf1**.
   The synthesized `<figcaption>` is added but the table's own
   `<table><caption>` is not cleared, so the text appears in both. Verified that
   `<table><caption>` is the normal non-float rendering, so this is specific to
   the float path. Q1's shape 1 carries the caption only in the figcaption.
   Also worth settling in that fix: the figcaption says
   `quarto-float-caption-bottom` while the inner table says `caption-top`.

2. **Figcaptions drop the "Figure 1: " / "Table 1: " prefix** — pre-existing,
   already filed as **bd-hb9a9ik8** (`prefix_caption` only handles `Paragraph`
   captions). The transform's unit tests assert the prefix is present at the AST
   level, so it is lost in the writer, not the transform.

Reviewer's call whether bd-4m2n6qf1 blocks this merge or lands as a follow-up.

## Remaining scope on the strand (not in this PR)

- `lst` e2e fixture exercising `#lst-` floats
- Preview inspection through the full WASM rebuild chain
- Layout panels (shape 3) — separate layout-engine sub-strand

Strand: bd-hcp8m3ve · Audit: bd-a0eyjshu


cscheid:
**Update on known issue 1 (bd-4m2n6qf1): checked against `main` — it is pre-existing, not a regression from this PR.**

Built `origin/main` (78d55deb) and rendered the identical minimal fixture:

```markdown
| a |
|---|
| 1 |

: Cap {#tbl-one}
```

`main` today:
```html
<div id="tbl-one">
<table class="caption-top table"><caption>Cap</caption>…</table>
Cap                                     <!-- bare trailing text node -->
</div>
```

this branch:
```html
<div id="tbl-one" class="quarto-float quarto-figure quarto-figure-center">
<figure class="quarto-float quarto-float-tbl">
<div aria-describedby="tbl-one-caption"><table class="caption-top table"><caption>Cap</caption>…</table></div>
<figcaption id="tbl-one-caption" …>Cap</figcaption>
</figure>
</div>
```

Both emit the caption twice. This PR changes the *shape* of the second copy — a stray trailing text node becomes a proper `<figcaption>` — without fixing the double emission. **So this is not a regression on that axis, and arguably an improvement.**

**Q1's fix, for reference.** It elides the table's own caption at float-*parse* time, in two places in `quarto-pre/parsefiguredivs.lua`:

- L280 (div-wrapped): hoists the table's caption into the float, then *"table caption should be removed from the table as we'll handle it"* → `table.caption = pandoc.Caption{}`
- L544 (caption-attr, `tbl-` identifier): *"we've parsed the caption, so we can remove it from the table"* → `el.caption.long = pandoc.Blocks({})`

Q2 builds its float DOM in the Finalization-phase transform rather than at parse time, so the natural Q2 site is wherever the figcaption is synthesized — clear the Table's own caption at the same moment.

**Suggested sequencing:** fix on top of this PR rather than against `main`. The bug predates the PR, but the PR rewrites `render_float_ref_target` — the exact code a fix must touch — so a fix landed on `main` would conflict. Either fold it in here or ship it as the immediate follow-up.

cscheid:
**Fixed in 32f06010 — folded into this PR as discussed.**

`render_float_ref_target` now elides a table float's own caption once it has been hoisted into the synthesized `<figcaption>`, matching Q1's behaviour (which does the same elision at float-*parse* time in `quarto-pre/parsefiguredivs.lua`).

Chose the transform over the writers deliberately: the caption is gone from the AST before either writer runs, so the pampa HTML writer **and** the React preview renderer both get the fix, with no render/preview parity drift.

Scoped to top-level `Table`s in the float content, and skipped when the float is uncaptioned — in that case nothing was hoisted, so the Table's caption is the only copy of that text.

**Before / after**, same minimal fixture (`: Cap {#tbl-one}`):

```html
<!-- before -->
<div aria-describedby="tbl-one-caption">
<table class="caption-top table"><caption>Cap</caption>…</table>
</div>
<figcaption id="tbl-one-caption" …>Cap</figcaption>
```
```html
<!-- after -->
<div aria-describedby="tbl-one-caption">
<table class="caption-top table">…</table>
</div>
<figcaption id="tbl-one-caption" …>Cap</figcaption>
```

Verified end-to-end through the binary on **both** repro forms (`: Cap {#tbl-one}` and `::: {#tbl-one}`) — "Cap" now appears exactly once in each.

**Worth noting for reviewers:** the existing `table_target_renders_q1_float_shape` supplies the caption as a *sibling paragraph*, never as the Table's own `caption.long` — so the duplicating path had no coverage at all. That is why 10511 passing tests missed this and only an e2e render caught it. The new test `table_float_clears_the_tables_own_caption` covers it, watched failing first.

`cargo xtask verify`: **all 14 steps pass**, 10512/10512 tests, zero snapshot churn.

Still open, deliberately not touched here: the figcaption carries `quarto-float-caption-bottom` while the inner table carries `caption-top`. That is a caption-location question, not a duplication one.

PR #424, opened 2026-07-27
hub-client: don't flash Offline before Online on open

### LINKED ISSUE #421: hub-client: documents always open showing initially as offline
Currently the timeout is set at just 1ms, so we start with the cached version, and show as offline.
Then when the connection happens, this gets updated. Often this is momentary, but this flash in status is not ideal.

shikokuchuo:
Closes #421.

Opening a document briefly showed the connection indicator as **Offline** before flipping to **Online**.

**Cause:** the indicator (`MinimalHeader`) mounts only after `connect()` resolves, so its first paint is whatever `connect()` decided. Production waited just **1 ms** for a peer, so `connect()` essentially always resolved offline-first — the header mounted Offline, and only the later `peer` event flipped it to Online.

**Fix:** widen the production peer wait 1 ms → **400 ms** (`PRODUCTION_PEER_TIMEOUT_MS`). `waitForPeer` resolves the instant the peer connects, so the common (fast) open now resolves as Online — the header mounts Online right away with the *live* document — adding only the real connect latency. A slower connect still mounts Offline and flips when the peer lands (rare tail, accepted — still better than always-first Offline). E2E path unchanged at 15000 ms.

Adds `MinimalHeader.test.tsx` covering the Online/Offline indicator (previously untested).

**Verified:** `tsc -b` clean, full hub-client suite 129 pass (incl. changelog qmd gate), `npm run build:all` green. Not verified in a live browser session.

### DIFF
--- a/claude-notes/plans/2026-07-27-connection-indicator-hide-until-resolved.md
+++ b/claude-notes/plans/2026-07-27-connection-indicator-hide-until-resolved.md
@@ -1,0 +1,66 @@
+# Connection indicator: no Offline flash on open (bd-53501yf7)
+
+## Overview
+
+When opening a document in hub-client, the header connection indicator
+**always showed "Offline" first, then flipped to "Online"** a moment later.
+
+### Root cause
+
+- `App.tsx` holds a boolean `isOnline`, initialized to `false`.
+- The indicator lives in `MinimalHeader`, which is inside `Editor` — and
+  `Editor` only mounts **after** `project` is set, i.e. **after**
+  `connect()` has already resolved. So the indicator's *first paint* is
+  whatever `connect()` decided; it never renders during the connecting
+  window, and the initial `false` is never seen.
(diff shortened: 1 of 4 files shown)


cscheid:
What's the badge displaying during those 400ms?

shikokuchuo:
The badge lives within `Editor`, which only renders once the `Project` is set, which happens after `connect()` has resolved - which is our up to 400ms wait.

cscheid:
(Clearly I'm not e2e testing this, so forgive my potential dumb questions)

> The badge lives within Editor, which only renders once the Project is set, which happens after connect() has resolved - which is our up to 400ms wait.

Ok, I think I see it. Do I understand correctly, then, that the behavior is: show "Online" as soon as we hear a reply from the sync server, but wait up to 400ms before claiming "Offline"?

shikokuchuo:
Exactly!

PR #432, opened 2026-07-29
Bring forward samod to 0.12.3

shikokuchuo:
Pull in our patched version of `samod` (rebased on 0.12.3) at `quarto-dev/samod@access-policy`.


cscheid:
@shikokuchuo we have a skill for checking multiple versions of a crate linked against q2, and our samod fork was holding back an upgrade of `rand`. Since you're in this part of the code base, can you take a look?

PR #441, opened 2026-07-30
Comments v1

vezwork:
This PR adds a default block render component for q2-preview and revealjs formats that 
- shows comment threads and allows you to add comments and resolve them.
- adds a toggle to the bottom bar that expands, collapses, and hides comment threads.

https://github.com/user-attachments/assets/62a318df-6709-431e-971a-7c2af087592c

## Notes
- I didn't try to get commenting on list table entries working, disabled that
- *commenting doesn't work on figures for some reason (it writes, but it causes broken syntax) so I've disabled commenting on figures.*
- comment bubbles try to position themselves at the top right of their associated block, but, especially when in expanded mode, bubbles can get large and its impossible for them to stay in that position without overlapping with other bubbles. To address this problem I've added force layout that ensures bubbles don't overlap. When you hover over the right half of a block, it highlights the associated comment bubble and moves the bubble to the top right of the block, pushing other bubbles around to make that happen.
- If you hover over the left half of a block, it doesn't cause any comment bubble interactions. I added this so people have an easy place to put their mouse while scrolling without causing bubbles to move around a bunch.

### DIFF
--- a/hub-client/q2-preview.html
+++ b/hub-client/q2-preview.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
@@ -12,11 +13,6 @@
       margin: 0;
       padding: 0;
     }
-    #root {
-      width: 100%;
-      height: 100vh;
-      overflow: auto;
-    }
   </style>
   <script>
     // Stub out React Fast Refresh runtime for iframe context
@@ -25,8 +21,10 @@
     window.$RefreshSig$ = function () { return function (type) { return type; }; };
   </script>
 </head>
+
 <body>
   <div id="root">Loading q2-preview renderer...</div>
   <script type="module" src="/src/components/render/q2-preview/entry.tsx"></script>
 </body>
+
 </html>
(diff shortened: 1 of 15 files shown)


hub-client/q2-preview.html
```diff
@@ -12,11 +13,6 @@
       margin: 0;
       padding: 0;
     }
-    #root {
-      width: 100%;
-      height: 100vh;
-      overflow: auto;
-    }
```
vezwork, line 19: this was causing there to be multiple scroll bars on the preview. It is necessary for revealjs tho, so that format adds this style programmatically now.

ts-packages/preview-renderer/src/q2-preview/PreviewRoot.tsx
```diff
@@ -1497,9 +1500,13 @@ export function PreviewRoot(props: PreviewRootProps) {
...
 
     // commitSubtreeEdit: send a subtree-channel PreviewNodeEditPayload (Plan 2b).
+    // The replacer strips every pool-index-carrying key: `s` (node
+    // SourceInfo), `a` (AttrSourceInfo), and `targetS` (Link/Image
+    // URL+title source refs — missing it caused InvalidSourceInfoRef
+    // on any subtree containing a link).
     const commitSubtreeEdit = (destinationSourceInfoJson: string, modifiedBlock: BlockNode) => {
         const stripped = JSON.parse(JSON.stringify(modifiedBlock, (key, value) =>
-            key === 's' || key === 'a' ? undefined : value,
+            key === 's' || key === 'a' || key === 'targetS' ? undefined : value,
```
vezwork, line 1509: don't know what is going on here, but I believe it fixed a bug where I could not edit or comment on anything that had a link in it.

cscheid (approved):
Aside from https://github.com/quarto-dev/q2/issues/442 (which we'll solve separately anyway), this looks great!

PR #453, opened 2026-08-05
hub-client: collections in project export/import + fix invisible import

icarusz:
## What

Two related fixes to the project-list export/import path, both found while testing the collections projects home against a fresh browser.

### Bug: imported projects didn't appear

"Import project list (JSON)" reported *"Imported 30 project(s)"* but displayed none. `importData` writes only the legacy IDB `projects` store, which the set-mode UI never renders; the IDB→set reconciler only ran on page load, so imports were invisible until a manual reload.

Fix: new `importProjectsAndReconcile()` (`projectSetReconciler.ts`) runs the reconcile inline, exposed as the `importProjects` action on `useCollectionSets` and wired through `ProjectsHome`. Imported projects appear immediately; result messages are honest about offline / already-present cases.

### Feature: collections in the export

Every export surface emitted a flat schemaVersion-4 list, so an export → import round trip restored projects but lost (a) collection groupings and (b) the collection doc pointers — no way to re-subscribe without hunting down invite links.

New `services/projectListExport.ts` defines the v5 format: projects plus collection pointers (docId, syncServer, display-only name/members/isRoot). Parse accepts v5, v4, and the pre-ExportData bare array; rejects newer-than-known versions. All three export surfaces (ProjectsHome, classic ProjectSelector, IDB-level `exportData`) include collections; import re-subscribes non-root collections via the existing `subscribeCollection` action. Import writes nothing into collection docs — membership and names arrive with sync.

## Testing

- 16 new tests (TDD, written failing-first) across `projectSetReconciler`, `projectListExport`, `projectStorage`
- Full `test:ci` green (876 + 109 + 130); `build:all` clean
- Verified end-to-end in local-prod twice (automated browser + manual Safari): export with a collection → fresh browser profile → import → "Imported N project(s), joined 1 collection(s)", collection restored with correct name and membership

Plan doc: `claude-notes/plans/2026-08-04-collections-export.md`


shikokuchuo:
@copilot resolve the merge conflicts in this pull request

shikokuchuo (approved):

PR #464, opened 2026-08-07
Live share for q2 preview

shikokuchuo:
## What this adds

`q2 preview` can now share a running preview session with other people.

- `q2 preview --share index.qmd` prints a join string.
- Someone else runs `q2 preview --join <string>` and opens the printed link. They see the same preview, live.
- The connection is end-to-end encrypted (iroh). The preview server still listens on loopback only, so the join string is the only way in.
- The new flag `--ui editor` opens the full hub-client editor instead of the viewer, for the host and guests alike.
- The new flag `--browser <name>` opens the session in a specific browser instead of the system default. It works for the host, `--share`, and `--join` alike, and conflicts with `--no-browser`.

## Who can do what

- Everyone can view the preview and re-run its code on the host machine.
- File edits reach the host's disk only when the host passes `--allow-edit`. The share banner says exactly what the join string grants.
- `--ui editor` without `--allow-edit` allows "ephemeral" editing: edits sync live to everyone in the session but never write to disk. The editor shows a persistent "Ephemeral session — edits won't be saved to disk" banner in this mode, for guests too (driven by `/api/preview/config`), so session-only edits cannot be missed.

## First-join payload

A guest always has the q2 binary installed, and that binary embeds the same preview UI bundles that the host serves. This branch uses that to cut the first-join download from the whole UI to a few kilobytes:

- The host advertises a SHA-256 manifest hash of its embedded UI assets in `/api/preview/config`. The guest compares it against its own during join setup, before the local port binds. On an exact match, the guest serves the UI from its own binary, and only the session's dynamic traffic (`/ws`, `/api/*`, `/auth/*`, `/health`) crosses the tunnel. On any mismatch (an older or newer q2, a dev build), everything tunnels as before: mixed-version joins still work, they only boot slower.
- The UI payload shrank regardless of serving mode. The hub WASM is 41.9 → 26.9 MB (release-profile tuning + `wasm-opt -Oz`). The embedded UI bundles now ship as identity-only **tar.zst archives** (~107 MiB → 17.0 MiB embedded), decompressed lazily on first asset request — `q2 render` never pays. Gzip responses are generated at runtime (level 9, cached per file) when `Accept-Encoding` allows, producing the same wire bytes the precompressed siblings did (~3.8× on the viewer dist). Content-hashed `/assets/*` now carry `immutable` cache headers. The `q2` binary is 41.6 MB smaller net from the WASM/dedupe work, and the archive embed takes a local release build from 193.7 → 106.6 MiB on top (−87.0 MiB, −44.9%).
- Measured through the real `--share`/`--join` binaries with a headless-Chromium boot driver. Wire payload: 54.67 MB → ~3 KB tunneled, plus ~10.8 MB served from the guest's own binary. First render on a simulated 10 Mbps / 100 ms link: **48.0 s → 1.5 s** (9.9 s for the all-tunnel fallback on the identical topology).
- Release CI compares the manifest hashes across all five per-platform artifacts and fails the release on drift. A mismatch is safe (guests tunnel), but it silently disables local serving for cross-platform pairs.

## How it works

- A new `quarto-p2p` crate provides the share tickets, the tunnel host (in front of the preview server's loopback port), and the tunnel client. The client (`TunnelClient::connect` + `bind`) serves a local proxy for the guest. All traffic goes over one encrypted iroh connection. If no relay is reachable, the banner says that guests can join over LAN/direct connections only.
- On a manifest hash match, the guest's `--join` runs a small L7 frontend on its local proxy port. A bounded head-peek (64 KiB / 5 s) routes each connection. The frontend answers exact manifest-hit `GET`/`HEAD` requests from the guest's own embedded bundle (byte-identical bytes and headers via the shared `asset_response` builder) and splices everything else onto the tunnel. `Connection: close` applies both ways: on local responses, and forced into tunneled non-upgrade heads. A browser cannot mix local and tunneled requests on one keep-alive connection. WebSocket upgrades pass byte-identical. There is deliberately no local index fallback: unknown paths always tunnel, so the host stays the single authority on what is dynamic. On a mismatch, `--join` bypasses the frontend entirely and acts as the plain thin proxy: the host serves everything through the tunnel. Host and guests share one Automerge document set either way.
- Editor-UI boots carry the share-route parameters (document id, file, project name) through `/api/preview/config`, so guests land in the same document instead of a setup screen.

## Fixes included

- **Sync clobber:** without `--allow-edit`, the hub's 5-second disk sync deleted any edit older than the previous sync tick, on the host and guests alike. The sync now skips its disk-merge step when the file on disk has not changed. Regression test included.
- **Guest-created files persist under `--allow-edit`:** previously, a guest's new file (New File dialog) or uploaded asset stayed VFS-only. The Automerge doc and index entry synced to the hub, but the hub skipped the missing disk file on every sync tick, so the file never landed on disk — contradicting the share banner's "EDIT the project's files on disk". The hub now distinguishes the two missing-file cases by sync checkpoint. No checkpoint means VFS-created: the hub writes the file to disk at the next sync tick (text and binary, parent directories created, symlink-escape paths rejected). A checkpoint means deleted on disk: the file keeps skipping and is never resurrected. Read-only sessions are unchanged. Six new hub sync tests cover this, plus an end-to-end run through the real `--share`/`--join` binaries with a browser-driven guest (create, upload, host edit back, delete stays deleted).
- Guests now boot straight into the editor share route.
- **No browser-storage accumulation from ephemeral sessions:** every preview session serves the editor from a fresh origin (a random loopback port), so the hub-client's IndexedDB writes — the `automerge` document cache and the `quarto-hub` project/identity/project-set records — were never read again and accumulated unboundedly across sessions. The preview-embed build now runs with `VITE_EPHEMERAL_STORAGE=1`: the automerge repos use the in-memory storage adapter (the same one the viewer SPA uses) and the hub records live in an in-memory facade behind the existing storage seam, so nothing persists per session. Reloading the editor rebuilds the session from the `/api/preview/config` `editorBoot` params instead of landing on "Project not found". This also retires the project-set migration-screen wart on fresh origins. The `quarto-cache` WASM artifact cache (SASS/theme artifacts) gets the same treatment: when the session is ephemeral it serves from an in-memory store with identical LRU semantics, in the viewer and editor alike — preview sessions leave no IndexedDB databases behind at all.
- The branch earlier carried a workaround for the Automerge `fork_at` collector panic (pinned fork + panic containment). That commit is gone: the branch is rebased onto main, and the samod 0.13.0 / automerge 3.4.1 bump (#513) fixes the panic upstream.

## Testing

- New integration tests cover share, join-through-tunnel, editor boot, and the sync clobber regression. Hub-client unit tests cover the banner's config fetch and validation, the in-memory storage facade and its consumers, and the `editorBoot` config parsing; the WASM bridge cache's in-memory mode has its own suite mirroring the IndexedDB contract (both run with no `indexedDB` global, so any leak into real IndexedDB throws). The ephemeral-storage behavior was verified end-to-end with real Chromium sessions against `q2 preview` in both editor and viewer modes: `indexedDB.databases()` is completely empty after boot and after a page reload, and the editor reload recovers the session.
- The join payload work was test-driven per phase: asset serving (gz negotiation, cache contract, HEAD semantics), manifest determinism and npm↔Rust generator agreement, the four mode decisions (match/mismatch/missing/override), `TunnelClient::connect` round-trip and token rejection, and 10 join-frontend tests (local serving with zero host asset requests, full-tunnel fallback, unknown-path tunneling, WS splice survival, 431 on oversize head, head-peek timeout, HEAD semantics, editor-UI boot).
- Real-browser end-to-end through the real `--share`/`--join` binaries: the guest serves all 12 boot asset requests locally, and only `/health`, `/api/*`, and `/ws` cross the tunnel (~3 KB). This e2e caught a bug that the per-connection tests could not: Chromium reuses idle *tunneled* keep-alive connections for later asset requests (~8.7 MB crossed the tunnel unrouted). The fix forces `Connection: close` into tunneled non-upgrade heads. Regression tests included.
- Cross-machine end-to-end run recorded: a GitHub Actions guest joined over the real relay.
- The archive embed is test-driven: new unit tests pin the identity-only embed (no `.gz` siblings in the binary), the runtime-gzip skip set sharing one source of truth with `precompress-dist.mjs` (`scripts/gzip-skip-extensions.txt`), and the gz→identity roundtrip; the existing asset-serving HTTP contract tests pass unchanged against the runtime-generated bytes.
- The full workspace Rust suite is green (11,939 passed), and `cargo xtask verify` is green, including the hub-client build and WASM tests.
- Re-verified after the rebase onto main (samod 0.13.0 / automerge 3.4.1): 905 Rust tests passed in the affected crates (1 skipped), and the WASM changelog-render gate passed (131 tests).
- Review follow-ups while draft: the ephemeral banner's config fetch now retries once when the first request fails, and the banner has its own component test. The `--join` help text mentions `--browser` and documents the embedded-asset serving and mismatch fallback.
- Fixed a CI flake: the two `share.rs` tunnel tests sometimes exceeded their 20s timeouts on the loaded ubuntu runner. They now run multi-threaded like every other tunnel test, with 60s caps.


vezwork:
whoa! iroh stuff!!? This sounds super awesome

vezwork:
I read through the `.ts` code a bit and I see there are barely any changes to the client. 

I was wondering if you've factored this PR such that we have an "iroh [automerge network adapter](https://automerge.org/docs/reference/repositories/networking/)"?

I found this existing [automerge](https://docs.iroh.computer/protocols/automerge) page in the iroh docs. It doesn't seem to be a network adapter? It seems very simple.

I don't really know whats going on but its cool.

shikokuchuo:
@vezwork it's not using any of the automerge-specific adaptors as our iroh tunnel is implemented lower level to tunnel all our http traffic - this includes several REST endpoints as well as the websocket connection. That's also why there's minimal change to the existing code.

cscheid:
... huh. One half of this session can even live on a pure webpage! <https://docs.iroh.computer/languages/wasm-browser>

shikokuchuo:
> ... huh. One half of this session can even live on a pure webpage! https://docs.iroh.computer/languages/wasm-browser

huh, I think you're right. So joining a session could theoretically not even need q2 installed!

PR #510, opened 2026-08-12
Fix indented continuation lines parsing as hard errors (bd-j7be7kuc)

cscheid:
## Summary

Fixes the P0 v0.18.0 regression tracked by `bd-indented-continuation-parse-error-j7be7kuc`: a soft-wrapped continuation line starting with `-`, `+`, or a digit became a **hard parse error when indented**, dropping the entire file from the render. Introduced by 92737cdd (the bd-w6tod0gh fix), whose new gate peeks were only exercised at indent 0. Real-world impact: the Posit Connect docs port lost its entire 53k-line API reference page.

A 280-cell characterization sweep (leader × indent 0–10 × context top/bullet/ordered/quote, pandoc 3.9 as reference) showed 118 regressed cells. All are fixed; the only remaining sweep errors are the 28 deliberate `*5 stars` unclosed-emphasis cells.

## Mechanism

When a SOFT_LINE_ENDING gate peek judged an indented line "prose", the peeked path skips `mark_end` (the lexer cannot rewind to post-indent/pre-delimiter), stranding the line's leading indentation. If the continuation resumed as plain text, the internal lexer produced `_whitespace` — which the grammar state after `_soft_line_ending` could not shift. Backtick/star peeks survived by accident (their continuations start with external-scanner tokens that absorb the whitespace); in-list cases survived only when the indent equaled the content column exactly.

## Fix (two parts)

- **Grammar**: `_soft_line_break` absorbs an optional trailing `_whitespace` under `prec.right`. Continuation-line leading whitespace contributes nothing to the inline stream, matching pandoc. Parser regenerated; error-message table regenerated (state renumbering).
- **Scanner**: the marker peeks are now shape-only; indentation judgment moved to the call sites where the reference column is knowable. Gate 1 (pre-`match_line`, raw columns) compares against a new `claimable_list_indentation()` (the leading LIST_ITEM* run of the block stack) + 3, skipping the peek on over-indent so the non-peeked `mark_end` absorbs the indentation; gate 2 (post-`match_line`) compares the residual indentation against 3 on both paths. This fixes three defects: the raw-vs-relative misread of `peek_ordered_marker`'s old internal guard (a nested `1.` at indent 4 now nests — the Connect-docs case), the missing guard on `peek_dash_plus_opens_block` (over-indented bullets are now lazy prose per CommonMark, not phantom block openers), and `first_peeked` being set by peeks that never advanced.

## Testing

- New 138-cell table-driven integration test (`crates/pampa/tests/integration/test_indented_continuation.rs`), fail-first verified: the 72+27+3 pre-fix failures matched the sweep prediction cell-for-cell. Expectations follow CommonMark; pandoc agrees on every regressed cell; deliberate qmd deviations are pinned as controls.
- New tree-sitter corpus file `test/corpus/indented-continuation.txt` (10 cases) adds the previously-missing indent dimension.
- `tree-sitter test` 582/582; `cargo nextest run --workspace` 11727/11727 with **zero snapshot changes**; full `cargo xtask verify` (including the WASM leg) green.
- E2E: the Connect-docs repro project renders **3/3 files (was 0/3)**; inspected HTML shows the digit-leading wrap joined into a single `<p>` and the nested `<ol>` markup present.

## Discovered work (filed, not fixed here)

- `bd-mt1ksg9b` — the `*` peeks share the same over-indent defect (pre-existing: `hello\n     * item` still hard-errors).
- `bd-z69hr4o0` — spurious `Space` after `SoftBreak` on indented backtick/star continuations (pre-existing).

Also adds `claude-notes/instructions/scanner-indentation-contexts.md`: scanner line-start changes must sweep indent × context, with the raw-vs-residual `s->indentation` trap and the `mark_end`/residual-whitespace mechanics documented.


cscheid:
Added 062f4f88: fixes a regression found while testing this PR against the Connect docs (nested ordered lists inside *attributed* fenced divs silently rendered as paragraph text — repro `attributed-div-ws-blank-list`). Root cause was in the scanner half of the fix, not the grammar: `claimable_list_indentation` now treats FENCED_DIV/FENCED_CODE_BLOCK/ANONYMOUS as transparent (they match without consuming a prefix), and the over-indent verdict uses the peeked skip-`mark_end` emission at both gates so the bogus paragraph-continuation fork dies across whitespace-only lines. The repro's variants now parse byte-identical to v0.18.0, and the div-wrapped nested-list case improves on 0.18 (real `<ol>` where 0.18 hard-errored). tree-sitter 582/582, workspace 11728/11728, zero snapshot churn, full `cargo xtask verify` green. Pre-existing Plain-vs-Para tightness quirk on whitespace-only separator lines filed as bd-25ezfgs8.

PR #544, opened 2026-08-18
fix(hub-client): restore Safari preview link clicks via allow-scripts + injected CSP

### LINKED ISSUE #128: hub-client: preview links fail in Safari
Clicking on links fails in Safari -> goes to a blank frame.

Example console log:

`Blocked script execution in 'https://quarto-hub.com/' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.`
`Refused to display 'https://quarto-hub.com/team-sync/2026-04-15.qmd' in a frame because it set 'X-Frame-Options' to 'DENY'.`

No issue in Chrome.

shikokuchuo:
## Fix preview links in Safari

Fixes #128.

### Problem

In Safari, a click on a link in the HTML preview opened a blank page in the preview frame.

### Root cause

The preview frame is sandboxed and did not include `allow-scripts`. WebKit blocks event listeners that the parent page attaches to a sandboxed frame ([WebKit bug 218086](https://bugs.webkit.org/show_bug.cgi?id=218086)). As a result, the click handler that intercepts links did not run in Safari. The preview frame followed the link itself. The hub server then refused to show the target page in a frame.

### The fix

- The sandbox now includes `allow-scripts`. The parent-attached listeners now run in WebKit.
- Each preview HTML payload gets an injected CSP meta tag: `script-src 'none'`. The tag blocks all scripts in the preview document: script elements, inline handlers, and `javascript:` URLs.
- The tag goes immediately after the `<!DOCTYPE>`. If the payload has no `<!DOCTYPE>`, the tag goes at the start of the payload.

### Security

`allow-scripts` plus `allow-same-origin` is a known sandbox-escape pair. The injected CSP tag is now the only blocker of scripts in the preview. If a preview document does not have the tag, a dev-mode warning fires. User content can add stricter CSP rules. User content cannot weaken this rule.

### Tests

- New unit and integration tests cover the injection and the sandbox attribute. They failed before the fix and pass after the fix.
- A new Playwright spec clicks a cross-document link in the preview. The spec reproduces the defect on real WebKit before the fix and passes after the fix.
- A second new Playwright spec proves that no script in preview content runs, on Chromium and on WebKit.
- Verified manually in real Safari: a click on a cross-document link in the preview now switches the editor to that file.
- CI now installs WebKit for these specs.
- Full `cargo xtask verify` passes.

### Not in this PR

- Follow-up strands: bd-ddfyqmfm (links to non-renderable files), bd-23es17uh (`frame-ancestors` header), bd-rpc3skz0 (Q2PreviewIframe inline handlers).

### DIFF
--- a/.github/workflows/hub-client-e2e.yml
+++ b/.github/workflows/hub-client-e2e.yml
@@ -158,11 +158,15 @@ jobs:
       - name: Pre-build hub binary
         run: cargo build --bin hub
 
-      # Install Playwright browsers
+      # Install Playwright browsers. Webkit is needed for the q2#128
+      # Safari regression specs (the webkit project in
+      # playwright.config.ts); the browser install adds ~1-2 min to CI
+      # even though test time stays flat (the project is testMatch-scoped).
       - name: Install Playwright
         run: |
           cd hub-client
           npx playwright install --with-deps chromium
+          npx playwright install --with-deps webkit
 
       # Delete all snapshots if recreating from scratch
       - name: Delete all snapshots (recreate mode)
--- a/claude-notes/plans/2026-08-17-safari-preview-links.md
+++ b/claude-notes/plans/2026-08-17-safari-preview-links.md
@@ -236,6 +236,10 @@ Rejected alternatives:
       on webkit-2287). The third-party-embed / PDF-`<object>` spot-checks
       and the `object-src 'none'` decision remain for user verification on
       production.**
+      → **2026-08-18 later: user verified manually in real Safari —
+      clicking a cross-document link in the preview now works. The
+      third-party-embed / PDF-`<object>` spot-checks (which gate the
+      optional `object-src 'none'` addition) remain open.**
 - [x] Note expected cosmetic change: Chrome may log an "allow-scripts +
       allow-same-origin can escape sandboxing" console warning (already present
       for the q2-preview iframe); script tags in rendered HTML now log CSP
--- a/claude-notes/plans/2026-08-17-safari-preview-links.md
+++ b/claude-notes/plans/2026-08-17-safari-preview-links.md
@@ -1,0 +1,297 @@
+# Fix quarto-dev/q2#128 — hub-client preview links fail in Safari
+
+## Overview
+
+Clicking links in the hub-client HTML preview fails in Safari: the preview iframe
+navigates away and goes blank. Console shows:
+
+1. `Blocked script execution in 'https://quarto-hub.com/' because the document's
+   frame is sandboxed and the 'allow-scripts' permission is not set.`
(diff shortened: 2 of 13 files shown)


cscheid:
> The sandbox now includes allow-scripts.

I... I think we need to be very careful here. Let's talk this through before we merge.

gordonwoodhull:
IMO we need to productionize @vezwork's `Q2SandboxedPreviewIframe` with the second domain for the react preview iframe, and also use the second domain for html preview iframe.

vezwork:
I refreshed myself on `Q2SandboxedPreviewIframe` and it does have `allow-scripts` and `allow-same-origin` (which is okay, because we will serve it from a different domain), see `Q2SandboxedPreviewIframe.tsx` . So this fix should be able to be ported to the sandboxed iframe project without issue. 

But:

> Each preview HTML payload gets an injected CSP meta tag: script-src 'none'. The tag blocks all scripts in the preview document: script elements, inline handlers, and javascript: URLs.

This is probably not sufficient for security because you can do a bunch of same-origin request stuff with just html. Also, we can't do this because our preview design includes running scripts (including user scripts) in the preview.

vezwork:
I read through the changes in the PR, and it does not change the sandboxing of q2-preview. It only changes DoubleBufferedIframe and MorphIframe:

- DoubleBufferedIframe is what we use for the default html format preview (not `q2-preview`)
- MorphIframe was an experimental alternative to DoubleBufferedIframe

This means this PR only fixes this bug for the default html renderer, which we are planning to move away from in favour of `q2-preview`.

vezwork:
And sorry, we need to actually move from `q2-preview` to q2 sandboxed preview at some point too 😅

shikokuchuo:
So the consensus seems to be that this won't work for q2 sandboxed preview - and we'll just have to fix it there in a different way when we get to it. Correct me if I'm wrong.

vezwork:
@shikokuchuo  It would be good to fix this issue in `q2-preview`. In productionizing the sandboxed preview, we will have to move all the functionality of `q2-preview` into the sandboxed preview (and its designed so that we can), and the fix would come along for the move easily. `q2-preview` has allow-scripts in its sandboxing, so adapting the fix in this PR to `q2-preview` shouldn't require any changes to the security sensitive stuff.

Though fixing this issue in `q2-preview` would require addressing the problem of clicking links at all (the rich and plain editors make it impossible currently). To address the problem of making links clickable in `q2-preview`, @cscheid suggested adding a toggle between "editing" and "viewing" preview modes next to the comment three-way-toggle in the bottom bar.

shikokuchuo:
Cool, thanks @vezwork  - I think "allow-scripts" in q2-sandboxed-preview should mean that we don't need what's in this PR, but I'll have another look, along with the mode toggle, next week!

PR #555, opened 2026-08-19
hub-client: deduplicate and regroup sidebar action buttons

shikokuchuo:
Stacked on the WCAG 2.2 PR (`feature/bd-trkzm9rq-hub-client-wcag-22`). Strands: bd-qhn2raky, bd-fj8sptb7.

## Changes

- **Project tab**: removed the "Choose New Project" button — the MinimalHeader switch-project button (top-left) calls the same `onDisconnect`, so the tab duplicated it. The unused `onChooseNewProject` prop is gone too.
- **Settings tab**: "Screenshot Preview" moved to the Project tab next to Export ZIP. It's an action that produces a download, not a preference; it also traded its one-off inline blue styling for the neutral outline used by secondary actions.
- **Settings tab**: the Editor/Preview section headings are gone — the four remaining toggles read as one flat list.
- **Files header**: New / Upload / Print are now icon-only buttons (file-plus, tray-arrow, printer — each with an aria-label and tooltip) that grow to equal widths and fill the row whether two or three are present. Print is last so the stable New/Upload pair never shifts when it appears. (Previously, three text buttons wrapped Upload onto a second row at the default 220px sidebar width.)

## Verification

- Tests written failing-first: restructured ProjectTab suite (prop removed; html2canvas mocked for the moved screenshot button), new SettingsTab suite pinning settings as preferences-only with no section headings, FileSidebar integration assertions for the icon buttons' accessible names.
- New committed e2e spec `e2e/files-header.spec.ts` asserts the buttons share one row at equal width and in stable left-to-right order in both the two- and three-button cases.
- `npm run test:ci` (unit + integration + wasm) and `npm run build:all` green; Playwright e2e 63 passed, 1 unrelated flaky in share-link-project-set (passed on retry).
- Real-browser visual checks with throwaway screenshot specs (deleted after): two-button row, three-button wrap regression, and the final icon row.

### DIFF
--- a/hub-client/changelog.md
+++ b/hub-client/changelog.md
@@ -25,6 +25,7 @@ WASM rebuild is needed for a changelog-only edit.
 
 ### 2026-08-19
 
+- [`b62746e7`](https://github.com/quarto-dev/q2/commits/b62746e7): Sidebar cleanup — the Project tab no longer duplicates the header's switch-project button, Screenshot Preview moves from Settings to the Project tab next to Export ZIP, the Settings tab drops its Editor/Preview section headings so the toggles read as one flat list, and the Files header's New, Upload, and Print actions become equal-width icon buttons that always share one row (Print last).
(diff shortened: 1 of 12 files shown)


shikokuchuo:
@vezwork To me these changes are self-evidently better relative to the status quo. I fully expect we'll move things around further.

PR #560, opened 2026-08-19
Record esbuild's @esbuild/* platform binaries in the lockfile (bd-y2z9w2wx)

cscheid:
## Summary

`package-lock.json` recorded top-level `esbuild@0.28.0` (a dependency of `@quarto/hub-mcp` and `@quarto/engine-host-deno`, added in the ts-engine-extensions merge) with **zero** `@esbuild/*` optional platform-dependency entries — only vite's nested, different-version esbuild copy had its platform set. Since esbuild's binary is delivered exclusively through those optional platform packages, every lockfile-driven install (`npm ci` in CI, incremental `npm install` locally) produced an esbuild JS wrapper with no executable, failing with:

```
Error: The package "@esbuild/<platform>" could not be found, and is needed by esbuild.
```

on **every** platform. First observed 2026-08-19 on darwin-arm64 when `cargo xtask verify`'s ts-packages leg hit the `quarto-hub-mcp` bundle test (`scripts/bundle.mjs`). `ts-test-suite.yml` (`npm ci` + `engine-host-deno` tests, which pin `esbuild@0.28.0`) is exposed to the same failure.

## The fix

Regenerate the lock entry: delete the stale `node_modules/esbuild` block (`jq 'del(.packages["node_modules/esbuild"])'`) and let `npm install` re-resolve it. The lockfile now records the full 26-entry `@esbuild/*` platform set; the diff is purely additive.

Also adds a troubleshooting note to `claude-notes/instructions/coding.md` documenting this class of problem (missing platform binaries for esbuild/rollup/swc/parcel) and the correct repair: fix and commit the **lockfile** — a machine-local `npm install --no-save @esbuild/<platform>` band-aid leaves CI and colleagues broken. This PR is deliberately separate from #559 so the incident has its own record should the class recur.

## Test plan

- Clean `npm ci` from the fixed lockfile, then `node -e "require('esbuild').buildSync(...)"` succeeds (previously threw the missing-package error).
- The previously failing `quarto-hub-mcp` bundle test now passes from that clean install: `npx vitest run src/bundle.test.ts` → 3/3.
- CI on this PR is the end-to-end check: `ts-test-suite` runs `npm ci` + esbuild-dependent tests on fresh runners.

### DIFF
--- a/claude-notes/instructions/coding.md
+++ b/claude-notes/instructions/coding.md
@@ -30,6 +30,8 @@ When making changes to `hub-client/`:
 
 1. **At the start of a dev session**, run `npm install` from the repo root to ensure all workspace dependencies are up to date. This is fast and prevents test failures from missing packages added by other contributors.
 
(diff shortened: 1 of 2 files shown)


cscheid:
@shikokuchuo, @gordonwoodhull: I wonder if we need stricter instructions/process in the repo for "new binary dependency", to avoid confusing the robots in situations like this. I've added some of them, but this is just a reminder for us to keep in mind that a new binary dependency (binaryen, esbuild are recent instances) changes the repo setup for everyone.

PR #565, opened 2026-08-19
hub-client: fix app stuck on an old version (GH #447)

### LINKED ISSUE #447: SPA can fail to update at login page
Repeated logins, even after leaving the page open for an extended period of time, kept producing the old SPA error message (the one from #435). This indicates that the new SW refresh was not being triggered.

Rendered moot for now by https://github.com/quarto-dev/q2/pull/446, but this still needs to be investigated.

shikokuchuo:
## Summary

Fixes #447. The deployed app could stay on an old version after reloads and new logins. Three causes, three fixes:

1. **The tab did not reload after an update.** The app now uses the `virtual:pwa-register` module. When a new service worker activates, a hidden tab reloads itself at once. A visible tab shows an "Update available" toast instead. That tab reloads when it becomes hidden, or when you click Reload.
2. **An idle tab never checked for updates.** The app now checks every hour, and each time a hidden tab becomes visible. Focus checks are throttled to one per 5 minutes so rapid tab switching stays cheap; the hourly check always runs on schedule because it drives the hidden-tab self-reload. A failed check — the machine is offline, say right after waking — is ignored quietly and retried on the next trigger. nginx serves `sw.js` with `no-cache`, so each check is a small conditional request.
3. **The install cache (precache) was too large.** It held about 56 MB, and one failed download discarded the whole update. The cache now holds only the app shell, about 14 MB. The WASM parser, the Monaco workers, and the sass compiler moved to on-demand caches (`wasm-cache`, `ondemand-assets`). `main.js` stays in the install cache, so the shell and the entry chunk always install together.

The service worker still activates updates at once (`registerType: 'autoUpdate'`). The plugin now stays registered in every build with `disable: isE2E || disablePwa`. E2E and preview-embed builds get a no-op stub.

## Measured

- Install cache: 169 entries / ~56 MB → 160 entries / ~14.3 MB
- `dist/sw.js` has no `.wasm`, `.worker-`, or `sass.default-` URLs in the precache list. Both on-demand routes are present.
- `dist/index.html` no longer injects `registerSW.js`.

## Verification

- `cargo xtask verify`: all 14 steps green.
- 967 unit tests and 131 wasm tests pass. 22 new tests pin the update policy and the toast.
- A Playwright smoke test passes. The E2E build emits no `sw.js`.
- A scripted browser run deployed a new build while a tab stayed open. The service worker claimed the page, `wasm-cache` filled on demand, the visible tab showed the toast, and Reload loaded the new bundle.
- Not verified in a browser: the hidden-tab self-reload. Headless Chromium cannot hide a tab. Unit tests cover the logic. Do a manual pass with `local-prod:nginx` before deploy.

## Follow-up (not in this PR)

After this change runs in production for some time, all old clients will have the new code. The server can then enforce nonces for all login tokens again (bd-mc00s2ws, left open by #446).


cscheid:
Exciting!

PR #589, opened 2026-08-24
Attribute render failures to the page that failed

gordonwoodhull:
A failed page now always names itself.

Two independent defects meant a page could die during a project render with the
`.qmd` name absent from the entire transcript, and — in one case — with a
confident diagnostic pointing at the wrong place.

## Engine-output parse errors are rendered against the wrong buffer

`EngineExecutionStage` parses the engine's executed markdown, so the resulting
diagnostic spans index into `result.markdown` — knitr's `.knit.md`, not the
author's `.qmd`. On success the stage registers that buffer as its own
`SourceContext` slot and remaps `FileId(0)` into it. On failure it returned a
bare `StageError`, dropping the context:

```rust
.map_err(|diagnostics| {
    PipelineError::stage_error_with_diagnostics(self.name(), diagnostics)
})?;
```

`run_pipeline`'s generic `StageError → QuartoError::Parse` conversion then
synthesizes a `SourceContext` from the *original document's* bytes and binds the
engine-output spans to it. For a 106-byte `knit.qmd` whose R chunk emits a bad
shortcode, that is a span at offset 4716 bound to a 106-byte buffer.

What the user saw depended on where the offset landed, and both outcomes were
bad.

**Past EOF** — the usual case, since engine output is much larger than its
source. `map_offset` fails, so ariadne renders nothing. The
`at <file>:<row>:<col>` fallback in `to_text_with_renderer` needs the same
resolution, so it is skipped too, and the file name goes with it:

```
Error [Q-2-34]: Unquoted shortcode parameter starting with digit
Shortcode parameter values starting with digits must be quoted.
```

**Inside the source file** — when the `.qmd` is long enough to contain the
offset. The error is then reported with a full frame at an unrelated location.
On a 14576-byte document this pointed at `big.qmd:68:100`, the letter `e` in a
filler paragraph with no shortcode anywhere near it. Nothing distinguishes that
from a real diagnostic.

This is the failure mode the `add-file-with-id` lint rule exists to prevent
(bd-m6wmztln): binding an assumed file to a diagnostic's resolved id renders
byte offsets against text they don't belong to.

The stage now binds those spans to the buffer they actually index and returns
`PipelineError::Structured`, which `run_pipeline` forwards untouched.

## The render summary drops the failing file's name

`print_render_diagnostics_text` prints `error: {input}: {error}` only for
`legacy_failures` — failures carrying no structured diagnostics. Everything else
goes through `coalesce_by_source`, and `CoalescedDiagnostic::to_text()` names
files purely through the diagnostic's own span, deliberately omitting the
`Affected files:` tail for single-file groups. So `FileFailure.input` — a
known-good path, sitting right there — was discarded, and a span-less diagnostic
printed with no file name at all:

```
Error: Execution failed in knitr: R process failed
```

`--json-errors` was never affected; it reads `input` structurally as
`source_file`. Only the text path lost it.

`failure_attribution_line` restores it. It asks whether the text about to be
printed names the page that failed, resolving the span the same way
`render_ariadne_source_context` does (`root_file_id`, then `map_offset(0, ctx)`),
and prefixes `error: while rendering <path>` when the answer is no. That covers
unresolvable spans and spans resolving to a file other than the one being
rendered. An ordinary parse error resolves to its own `.qmd`, so existing output
is byte-identical and no snapshots change. Path matching is separator-insensitive:
the intermediate goes through `to_forward_slashes` while `FileFailure.input`
stays native, so a raw `Path` comparison would miss on Windows and add a
redundant line to every parse error there.

## Result

```
error: while rendering /…/knit.qmd
Error: [Q-2-34] Unquoted shortcode parameter starting with digit
     ╭─[ /…/knit.knitr.rmarkdown:138:23 ]
     │
 138 │ {{< fa envelope size=1x >}}
     │                       ┬
     │                       ╰── Shortcode parameter values starting with digits must be quoted.
─────╯
```

Both the page and the offending line are named, and the frame shows real text
from the buffer the error came from.

The motivating case was `download.qmd` in the Positron docs port — one of four
pages still failing to render there, and the most linked-to page on the site
(83 of 89 rendered pages reference `download.html`). With no file name in the
output it was read as a freeze problem for several days.

## Not addressed

Line numbers still refer to the engine intermediate rather than the source.
knitr's traceback continues to say `failing.rmarkdown:133-135`, and the frame
above is anchored at `knit.knitr.rmarkdown:138`. For content the engine
*generated* there is no `.qmd` line to point at, but for content it merely
passed through — a shortcode the author typed — the reported line is unhelpful.
Fixing it needs a line map from the serialized intermediate plus per-engine
stderr parsing. Tracked as **bd-1ng2ivis**; `serialize_ast_to_qmd` already
computes a `SourceInfo` that may be most of the raw material.

`crates/quarto-core/src/engine/preview_record.rs:241` uses the same
`stage_error_with_diagnostics` shape and may share the defect. It feeds
q2-preview rather than the CLI summary and is unexamined here.
`serialize_ast_to_qmd`, the other sibling site, is fine — its spans index the
original AST's source, so the generic conversion binds them correctly.

## Testing

Ten new tests. Two drive `EngineExecutionStage` with a fake `ExecutionEngine`
that emits unparseable markdown, asserting the error carries a context in which
the span resolves, and that the span falls inside the engine output rather than
past the end of the source. Six cover `failure_attribution_line` directly:
location-less diagnostics, spans resolving to the failing page, spans resolving
to an engine intermediate, spans past EOF, multi-file groups, and mixed path
separators. Two exercise the real binary end to end, gated on `Rscript` plus the
`knitr` package following the pattern in `marimo_engine_e2e.rs`.

Workspace: 13140 passed, 199 skipped, against a baseline of 13130. The
baseline's single failure, `quarto-core
engine::ts_engine::tests::test_race_free_instance_exclusive`, is a pre-existing
flake — it times out at 15s under full contention and passes in 0.3s alone.

Closes bd-render-failure-unattributed-yxe0v7th.

### DIFF
--- a/claude-notes/plans/2026-08-24-render-failure-unattributed.md
+++ b/claude-notes/plans/2026-08-24-render-failure-unattributed.md
@@ -1,0 +1,223 @@
+# Render failures with no attribution (bd-render-failure-unattributed-yxe0v7th)
+
+## Overview
+
+A page can fail during a project render and the output never names the file.
+The strand reports two shapes; investigation confirmed both and found a third,
+worse face of shape 1.
+
+### Shape 1 — engine-output parse errors are bound to the wrong buffer
+
+`EngineExecutionStage` parses the engine's executed markdown
+(`result.markdown`, e.g. knitr's `.knit.md`) at
+`crates/quarto-core/src/stage/stages/engine_execution.rs:611-621`. On the
+**Ok** path it carefully registers the intermediate as a new `SourceContext`
+slot and remaps `FileId(0)` into it. On the **Err** path it throws that
+context away:
+
+```rust
+.map_err(|diagnostics| {
+    PipelineError::stage_error_with_diagnostics(self.name(), diagnostics)
+})?;
+```
+
+`pampa::readers::qmd::read` returns only `Vec<DiagnosticMessage>` on failure,
+so the `SourceContext` that gave those spans meaning is dropped. The generic
+`StageError -> QuartoError::Parse` conversion at
+`crates/quarto-core/src/pipeline.rs:725-741` then **fabricates** a
+`SourceContext` from the *original document's* bytes and binds the
+engine-output spans to it.
+
+Measured, `q2 render` on a 106-byte `knit.qmd` whose R chunk emits a bad
+shortcode:
+
+```
+DBG conv: source_name=.../knit.qmd content_len=106
+          diags=[("Unquoted shortcode parameter starting with digit",
+                  Some((4716, "Original { file_id: FileId(0),
+                               start_offset: 4716, end_offset: 4717 }")))]
+```
+
+Offset 4716 indexes the `.knit.md`; the bound buffer is 106 bytes. Two
+outcomes follow from where the offset lands:
+
+- **Past EOF** (the common case — engine output is much larger than the
+  source): `map_offset` fails, so ariadne renders nothing *and*
+  `to_text_with_renderer`'s `at <file>:<row>:<col>` fallback is skipped too.
+  Output is a bare title + problem with no file, line, or frame. This is the
+  shape the strand reports.
+- **Inside the source file** (a long `.qmd`): a fully-formed, confident
+  ariadne frame is printed against an **arbitrary wrong location**. Measured
+  on a 14576-byte `big.qmd`: the error was reported at `big.qmd:68:100`,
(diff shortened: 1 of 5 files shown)


gordonwoodhull:
TL;DR print errors pointing at intermediate markdown output by engines. (which will later be enriched to point to source too.)

PR #628, opened 2026-08-27
Feature/project and document bars split

vezwork:
Screenshot 2026-08-28 at 1 20 28 PM

Screenshot 2026-08-28 at 1 21 40 PM


shikokuchuo (commented):
Nice, definitely with you on paring back what we have around the edges, so the page itself stands out more.

1. Agree on the preview button design
2. Suggest you rebase and pick up the sidebar toggle button
3. Perhaps think about being less boxy - I feel the lines and in particular how they cut and divide things feels a bit rigid
4. Just one synced time please (Two is not actionable information - what do I do if one is connected and one not? What does it mean that one synced time is behind the other? Suggest combining into one user-facing measure)

shikokuchuo:
@vezwork I just finished merging a bunch of accessibility work. Sorry if it's causing noise for you, but if you rebase to main all should be good!

vezwork:
image

@shikokuchuo here's an update.

> 1. Agree on the preview button design
> 2. Suggest you rebase and pick up the sidebar toggle button
> 3. Perhaps think about being less boxy - I feel the lines and in particular how they cut and divide things feels a bit rigid
> 4. Just one synced time please (Two is not actionable information - what do I do if one is connected and one not? What does it mean that one synced time is behind the other? Suggest combining into one user-facing measure)
Re:
1. nice
2. ✅
3. tried to add more roundedness, lmk what you think
4. (copied and pasted from Slack:) we've had a bunch of issues with confusion (and bugs? hard to tell) around the sync status being different between the project and the document. I think its important to show both (at least for now, until we come up with a better way to think about this with exposing as much info to the user). This was actually the main change I wanted to make lol, I just accidentally did a big reorganzation to accomadate it.

shikokuchuo:
> 3. tried to add more roundedness, lmk what you think

Liking this.

> 4. (copied and pasted from Slack:) we've had a bunch of issues with confusion (and bugs? hard to tell) around the sync status being different between the project and the document. I think its important to show both (at least for now, until we come up with a better way to think about this with exposing as much info to the user). This was actually the main change I wanted to make lol, I just accidentally did a big reorganzation to accomadate it.

The way you have it in the screenshot is more informational and not in your face, such that I don't seem to mind. But if the goal is for diagnostics, then I think we should figure out what's actually useful and just show that. I'm fine to show both in the meantime in any case, but remember ultimately this is not an affordance for us.

PR #629, opened 2026-08-27
Fix sync-client test mocks missing repo.handles (TS suite red on main)

shikokuchuo:
## Summary

The TS Test Suite `workspace-ts-suites` job has been red on main since 5d31edf6b ("Add connection modal and fix small isOnline bug") — 29 failing tests across three quarto-sync-client files, all with `TypeError: Cannot convert undefined or null to object`.

## Root cause

5d31edf6b added a `trackPeers` handle-sweep to `client.ts` that reads `repo.handles` (a documented, always-initialized getter on the real automerge `Repo`) and re-sweeps on a 2s `setInterval`. The six `Repo` mock implementations in the test files were never updated:

- `Object.values(repo.handles)` threw in every test using those mocks (9/9 in `applyEditorOperations.test.ts`, 9/9 in `client.test.ts`, 11/12 in `client.connect-options.test.ts`).
- Two `connect-options` tests asserted exact fake-timer counts (`toBe(0)` / `toBe(1)`) that the new always-on watch interval broke.

## Fix (test-only, 18 insertions / 4 deletions)

- Add `handles: {}` to the six `Repo` mock implementations — fixing the mocks rather than making production code defensive, since the real `Repo` guarantees the property.
- Update the two timer-count assertions to account for the watch interval, with comments explaining the timer breakdown.

## Verification

- `npm test -w ts-packages/quarto-sync-client -- --retry=2` (the exact CI step): **137/137 tests pass across 21 files** (was 29 failures).
- `tsc --noEmit` clean.
- No other CI leg affected: test-only change, no `src/` or dist impact.

Tracked as bd-iscjjhx8.

### DIFF
--- a/ts-packages/quarto-sync-client/src/applyEditorOperations.test.ts
+++ b/ts-packages/quarto-sync-client/src/applyEditorOperations.test.ts
@@ -127,6 +127,8 @@ async function setupClientWithFile(initialText: string) {
       import: vi.fn().mockReturnValue(indexHandle),
       create: vi.fn().mockReturnValue(fileHandle),
       networkSubsystem: mockNetworkSubsystem,
+      // client.ts trackPeers() sweeps repo.handles on connect
+      handles: {},
     });
     return this as Repo;
   } as unknown as typeof Repo);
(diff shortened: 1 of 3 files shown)


shikokuchuo:
@vezwork seems your earlier commit required some test changes. This PR fixes those.

PR #640, opened 2026-09-02
Diagnose a missing shortcode separator as Q-2-52

gordonwoodhull:
A shortcode's delimiters take a space on the inside: `{{< fa plus >}}`. Written tight, the diagnostics were wrong in three different ways.

| input | before | after |
|---|---|---|
| `{{<fa plus>}}` | fatal, **uncoded** | `[Q-2-52]` |
| `{{<fa plus >}}` | fatal, **uncoded** | `[Q-2-52]` |
| `{{< fa plus>}}` | `[Q-2-34]` unquoted parameter starting with digit | `[Q-2-52]` |
| `{{< fa plus >}}` | parses | parses |
| `{{< fa 2plus >}}` | `[Q-2-34]` | `[Q-2-34]` |

The uncoded message named no construct and offered no remedy. The Q-2-34 case was worse: the parameter is `plus`, there is no digit on the line, and doing what the message says — quoting the value — reaches a *different* uncoded state, so following the advice moved you from a wrong code to none.

It also cascaded. One missing space five lines above a valid callout produced three errors, the third blaming the callout's closing `:::`:

```
Click the {{<fa plus>}} icon.        <- error 1
                                     <- error 2, on the `}}`
::: {.callout-note}
A callout well after the bad shortcode.
:::                                  <- error 3
```

Requiring the space stays. `_shortcode_sep` has no empty alternative, which is what keeps `{{` unambiguous with Jinja, Handlebars and the other template syntaxes Quarto documents quote. The grammar is unchanged.

## Q-2-52

```
Error: [Q-2-52] Shortcode delimiter not separated by a space
   ╭─[ cascade.qmd:5:14 ]
   │
 5 │ Click the {{<fa plus>}} icon.
   │              ─┬
   │               ╰── A shortcode's delimiters must be separated from its
   │                   content by whitespace. Write `{{< fa plus >}}` — a space
   │                   after `{{<` and a space before `>}}` — rather than
   │                   `{{<fa plus>}}`.
───╯
ℹ Both delimiters need the space. Adding it to only one of them leaves the
  shortcode unparseable.
ℹ Quarto 1 accepted the tight spelling because it matched shortcodes with a
  regular expression. Quarto 2 parses them.
ℹ `qmd-syntax-helper convert -r q-2-52 -i <file>` rewrites both delimiters.
ℹ Parsing could not continue reliably past this point, so any further errors
  in this file were not reported. Fix this one and try again.
```

The cascade above is now one error, on line 5.

## Two additions to the error table

The table is keyed on `(LR state, lookahead symbol)`, and the tight closing delimiter reaches exactly the state Q-2-34's three corpus cases reach. Q-2-34 was not matching a delimiter — it was the only entry at that key and answered by default. Two mechanisms make room for a second entry there, and stop the cascade:

**`guard`** — a corpus case may carry a regex matched against the source from the error position to the end of its line. Selection is by specificity: entries whose guard rejects the text drop out, and if any guarded entry survives, the unguarded ones drop out too. An uncompilable pattern admits nothing, so a corpus typo means "never fires" rather than "always fires". Two entries use it — the tight closing delimiter, which claims its shared state only when the error is *at* the delimiter, and the escaped keyword-value form, which shares a state with Q-2-28's line-break spelling. A guard is per case, never per code: one code covers several spellings failing in different states, and usually only one of them is contended.

**`desynchronizes`** — a code may declare that this error leaves the parser unable to resynchronise, in which case diagnostics positioned after the first such error are dropped and the survivor says so. Errors before it are kept. This is not a heuristic for the shortcode case: recovery from the tight `{{<` consumes the callout's opening `:::` and reads it as three `pipe_table_align_left` tokens, which is why the closing `:::` reads as unmatched.

Both fields are optional and default off, so every existing corpus file and table entry is unaffected.

## Coverage

Q-2-52 spans 75 parser states, because the two delimiters fail in opposite places. The opening delimiter is consumed in the *enclosing* inline container's state, so it needs a case per container: emphasis and strong in both spellings, strikeout, super- and subscript, both quote spellings, inline notes, editorial spans, link text, image alt text, bracketed spans, and pipe-table header and body cells. The closing delimiter is reached from inside the shortcode, so it needs one case per argument shape instead — bare name, naked value, number, single- and double-quoted values (one-character content lexes differently from longer), each of those as a keyword value, values the language-specifier scanner does not claim (non-ASCII, or punctuation-initial), and a nested shortcode — except after a bare name, where the parser has not committed to an argument list and the container shows through again. Attribute values and link destinations have their own states. The escaped spelling `{{{< … >}}}` has its own families throughout.

A 17 × 44 matrix over containers and spellings pins all of it, and asserts alongside that the corrected form of every row parses in every container.

Two nearby failures are deliberately not claimed, with tests saying so: a positional argument after a keyword argument, and an escaped shortcode used as another shortcode's argument. Both fail with the space as well as without it.

Q-2-34 keeps its rule and its cases. `{{< fa 2plus >}}` still reports Q-2-34, and so does `{{< fa 2plus>}}`, where both faults are present and the digit-led value is the earlier one.

## Two things the corpus needed but did not have

**Nothing reconciled "the case that defines a code" against "the code that case produces."** The existing corpus tests check that a diagnostic is *rendered* — box-drawing characters, a resolvable location — not that it is the right one. A new entry could quietly claim a state belonging to an existing code and the suite would stay green, which is the exact defect this corpus exists to prevent. `every_corpus_case_reports_its_own_code` now checks all 784 case files.

**That test could not have worked as written.** `include_error_table!` read the generated table with `fs::read_to_string`. A proc macro reading a file that way leaves no trace cargo can see, so editing the corpus alone never marked pampa dirty and the previously embedded table kept being used. Mutation-testing the new test found this: removing both `guard` fields left it passing. The macro now emits an `include_str!` so rustc tracks the file, and the same mutation fails 8 cases. Filed as bd-gfov647r, since the window in which corpus edits could have been validated against a stale table goes back to whenever that macro was written.

## Note on the docs page

`docs/errors/markdown/Q-2-52.qmd` writes its bad examples as prose with escaped braces rather than in code blocks. Shortcode syntax inside a code span or a fenced block is currently expanded rather than shown literally — `{{< fa plus >}}` renders as `?fa` — which is a separate bug, filed as bd-xfqx2tuc. The correct spellings use the escaped `{{{< … >}}}` form, which is the existing convention in `docs/guides/authoring/shortcodes.qmd`.


cscheid:
`every_corpus_case_reports_its_own_code` is a very nice addition. I had noticed this in the past but needed to move on to work on other things.

PR #646, opened 2026-09-03
Emit block-level raw HTML as a block instead of wrapping it in a paragraph

gordonwoodhull:
A line beginning with a block-level HTML tag was parsed as inline content, so it landed inside a paragraph and the rendered page carried a spurious `<p>` wrapper:

```html
<p><details class="checksums-section">
<summary>SHA-256 checksums</summary></p>
<p>Text inside the details.</p>
<p></details></p>
```

That is invalid HTML. `<details>` and `<div>` are flow content, not phrasing content, so a browser's HTML5 parser force-closes the `<p>` and reparents everything after it. The DOM the page ends up with is not the one the markup describes, which changes what descendant and child selectors match. The visible text is identical either way, so the breakage never showed up in a text-level comparison — it surfaced on the Positron docs' `download.qmd`, where a tag-stack HTML comparator lost track of `<main>` and reclassified the entire SHA-256 checksums table as page chrome.

A paragraph whose first inline is a block-level raw HTML tag is now emitted as a `RawBlock`. The extent needs no adjustment: a markdown paragraph and a CommonMark type-6 HTML block both run to the next blank line, so lifting the whole paragraph reproduces the HTML block exactly — including the two-line `<details>` / `<summary>` shape above, which becomes a single `RawBlock`. On the reported fixture the output is now byte-identical to Quarto 1's and to bare pandoc 3.8.1's, and `download.qmd` renders `<details>` at block level with the checksums table inside it and no paragraph wrappers anywhere.

Naked HTML is still not a supported authoring form — `crates/pampa/README.md` lists it among the intentional divergences, and `dev-docs/syntax-notes.md` rejects pandoc's interleaving approach because it leans on parser-combinator backtracking. `Q-2-9` still fires on a lifted block. This makes the fallback emit valid HTML; it does not turn naked HTML into a feature. The documented spellings, a `{=html}` fence and a `::: {.class}` div, were already correct and are unchanged.

Which tags lift comes from pandoc's block-tag whitelist rather than CommonMark's start conditions. CommonMark's HTML block type 7 promotes any complete tag alone on a line, which would newly diverge from Quarto 1 for a lone ``, `<a href=…>` or `<span class=…>` — common shapes, and a worse trade than the bug being fixed. pandoc's DocBook names and the type-6 names pandoc lacks are excluded too; `html_block.rs` carries the reasoning for each exclusion. `native_divs` and `native_spans` are not ported: they require balanced tag matching across arbitrary block content, and the HTML output is identical without them.

`Q-2-9`'s message was inline-only in both halves — it claimed a `RawInline` and suggested `` `<element>`{=html} ``, neither of which is right for a block. It now branches, pointing block-level authors at a `{=html}` fence or a Quarto div. The error docs page is rewritten to match. The qmd writer's `RawBlock` path also picks up the HTML-comment case that the inline writer already had, so a standalone comment keeps round-tripping as `<!-- … -->` rather than degrading to a fence.

Two gaps are deliberate. A block-level tag in the middle of a paragraph does not interrupt it, and a tight `<div>` / `</div>` with no blank lines stays one verbatim `RawBlock`, so markdown inside it is not processed. Both disappear with a blank line after the opening tag, which is the shape in the fixture and in `download.qmd`.

## Testing

`cargo nextest run --workspace` — 13624 passed, 0 failed, 199 skipped, against a 13590 baseline; the +34 is exactly the tests added here. Clippy, `cargo xtask lint` and `cargo xtask verify --skip-hub-build` are clean.

## Snapshots

27 modified, all `crates/pampa/snapshots/json/html-comment-*`, all the same change — a standalone HTML comment moves from `Para[RawInline]` to `RawBlock`. 26 are top level; the 27th places the `RawBlock` inside a `BlockQuote` with no `> ` prefix in its text, matching pandoc byte for byte. No other snapshot in the workspace changed.

## Related

Closes bd-block-html-wrapped-in-p-w8qebxig.

Filed separately while working on this: bd-xmw5ekr6 — `<!DOCTYPE html>` and `<?xml ?>` at line start are a parse error on released 0.28.0 and on main, because the scanner routes a leading `<!` into comment lexing. That is upstream of this change, so the lift cannot reach it; `starts_block_html` already returns true for both, so nothing here needs changing once the scanner emits them.


cscheid:
It's a good kindness to fix this!

But just for future reference, I think we want to nudge users away from naked raw HTML in qmd files anyway.

PR #647, opened 2026-09-03
Don't render a listing into the section that shares its id

gordonwoodhull:
A heading with no explicit id gets one derived from its text, so `## Learn more` becomes `<section id="learn-more">`. Naming a listing after the section it sits in is a natural thing to write:

````markdown
---
listing:
  - id: learn-more
    contents: posts/*.qmd
---

## Learn more

::: {#learn-more}
:::
````

Doing that destroyed the section. The slot search walks the document outside-in for a Div carrying the listing's id, and by the time listings render, `SectionizeTransform` has already wrapped the heading in a Div carrying that same auto-generated id. The section matched before the author's `:::` slot nested inside it, and its entire content was overwritten with the listing — the heading and every paragraph in that section gone from the output. Only content outside the section survived. The render exited 0 and emitted nothing.

A section is now never a listing host. The search recurses past it to the author's real slot; when the section holds no slot at all, the listing falls through to the append path and gets its own container. Either way the section keeps its heading and its prose, which is what Quarto 1 does.

That still leaves the author with a heading and a listing claiming the same `#learn-more` anchor, which is worth saying out loud, so the collision now warns:

```
Warning [Q-12-25]: Listing `learn-more` has the same id as the section for heading
“Learn more”. The section keeps its content and the listing renders separately, but
both elements claim the `#learn-more` anchor, so links to it are ambiguous. Rename
the listing's `id:` (along with any `::: {#learn-more}` slot that matches it), or
give the heading an explicit id of its own.
```

The warning fires on the pass that renders, not on every pass over an already-rendered document.

Before and after on `docs-quarto-2/assistant.qmd`, which has exactly this shape:

```html
<!-- before -->
<section id="learn-more" class="section level2" data-listing-rendered="1">
<div class="custom-card-grid">...

<!-- after -->
<section id="learn-more" class="section level2">
<h2>Learn more</h2>
<div id="learn-more" data-listing-rendered="1">
<div class="custom-card-grid">...
```

On the `section` class as the discriminator, rather than preferring the innermost match: innermost still selects the section when the section contains no `:::` slot, so that variant of the bug would survive. Rejecting hosts whose id came from a heading's auto-id would still allow the destruction when someone writes `## Learn more {#learn-more}` explicitly. `section` is the class the HTML writer keys on to emit `<section>` instead of `<div>`, so the rule says precisely what is meant: a listing never renders into something that will become a `<section>`.

Fixes bd-listing-id-collides-with-heading-l57w41jl.

### DIFF
--- a/crates/quarto-core/src/transforms/listing_render.rs
+++ b/crates/quarto-core/src/transforms/listing_render.rs
@@ -276,7 +276,37 @@ fn render_one(
     };
 
     // Splice into the AST.
-    if !try_replace_explicit_slot(ast, &r.listing.id, &parsed_blocks) {
+    let outcome = try_replace_explicit_slot(ast, &r.listing.id, &parsed_blocks);
+
+    // A section is never the host (see `fill_in_blocks`), so the
+    // section's content is safe either way — but the author still
+    // has a heading and a listing competing for one HTML anchor,
+    // and only they can decide which should keep the id. Report it
+    // on the pass that actually renders, so a second pass over an
+    // already-rendered document stays quiet.
+    if outcome != SlotOutcome::AlreadyRendered
+        && let Some(heading) = find_colliding_section(&ast.blocks, &r.listing.id)
+    {
+        let which = if heading.is_empty() {
+            "a section".to_string()
+        } else {
+            format!("the section for heading \u{201c}{heading}\u{201d}")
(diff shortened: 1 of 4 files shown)


cscheid:
This is a really good fix.

PR #650, opened 2026-09-03
Don't execute a cell that is being displayed (bd-knitr-executes-nested-display-fence-atbtktdj)

gordonwoodhull:
## The bug

Consider a page that both *runs* an executable code block and *shows* one:

`````qmd
---
title: Using R in Quarto
engine: knitr
---

Summing the first ten integers gives:

```{r}
sum(1:10)
```

To write a cell like that yourself, wrap it in a longer fence so it is
displayed rather than run:

````markdown
```{r}
sum(1:10)
```
````
`````

A four-backtick markdown display block whose content is a three-backtick `{r}` fence gets executed by knitr, and the displayed example is *replaced* by knitr's own cell scaffolding. The author writes ```` ```{r} ````; the reader is shown `::: {.cell}` and `{.r .cell-code}`.

It is reachable only when the engine is already live. With no real cell in the document, AST-based resolution correctly declines to start knitr — so one added real cell anywhere turns every displayed example on the page into a live cell.

Two symptoms are worse than "wrong output":

- **A displayed `{python}` example in an R document kills the render.** knitr hands the unclaimed language to reticulate: `python_not_found`, Execution halted. This is the most user-visible face of the defect, and it settles the language predicate — masking only the current engine's languages would not fix it.
- **A blockquoted display block loses its contents entirely,** rendering empty. knitr's own pattern is `^[\t >]*`, so those fences really do execute.

Inherited from Quarto 1; not a q2 regression. Sibling engines differ: the Julia engine gets this right (QuartoNotebookRunner parses in Julia, nesting- and indentation-correct) and marimo gets the top-level case right via `breakQuartoMd`'s length guard.

The strand was filed **wontfix-candidate**, on the grounds that q2 "has no seam short of escaping nested fences before hand-off, which would corrupt the display content a different way." This PR is that seam, with the corruption avoided by a byte-exact round trip.

## The fix

One seam in `engine_execution.rs`, inside the per-engine loop:

```
mask -> serialize_ast_to_qmd -> engine.execute -> unmask -> capture -> reparse -> reconcile
```

Mask rewrites ```` ```{r} ```` to ```` ```{.r q2-nested-executable} ````. **The leading dot is what does the work** — all three partitioners that could execute the fence require an alphanumeric first character after `{`:

| scanner | pattern fragment |
| --- | --- |
| knitr `all_patterns$md$chunk.begin` | `\{([a-zA-Z0-9_]+( *[ ,].*)?)\}` |
| q2 jupyter `parse_code_blocks` | `\{(\w+)\}` |
| TS `breakQuartoMd` | `\{([=A-Za-z][=A-Za-z0-9._]*)...\}` |

The marker class exists only so `unmask` knows which openers were ours, and is never produced by an author writing their own `{.lang}`.

Four things worth a reviewer's attention:

**It must be inside the loop.** `to_run` is iterated with the AST re-serialized each time and left unmasked between iterations, so masking once beforehand would leave a second engine unprotected. Because every engine reaches this one serialize call, knitr, q2's jupyter text engine, and TS extensions are all covered with no per-engine work.

**Prefix handling is asymmetric on purpose.** `mask` sees a `CodeBlock`'s text, from which the reader has already stripped any blockquote `> `; only `unmask`, running textually over engine output, ever meets one. So mask needs no prefix logic and unmask must not be `^`-anchored.

**Byte-exactness is load-bearing, not cosmetic.** Reconcile compares against the unmasked original. One byte of drift and it treats the display block as *changed* and replaces it — handing the author's example the `.rmarkdown` intermediate's `source_info`. A test asserts the block lands in `blocks_kept` with its original `source_info`, not merely that its text looks right.

**The capture is asymmetric by necessity.** `input_qmd` stays masked (ReplayEngine byte-compares it against a live input that is itself masked); `result.markdown` is unmasked before the emit, or `q2 preview` renders the marker. `compute_input_qmd` has three consumers that want different bytes — the staleness compare and cache key need masked, while `write_review_file` is security-facing and unmasks, because an operator deciding whether to permit execution must see the `{r}` they actually wrote.

## Provenance

Masked text is longer than its source. The qmd writer's map is output-indexed, so a lengthened block shifts nothing for any *other* block — but within it, `Concat::map_offset`'s `Original` arm computes `start_offset + offset` with no clamp to `end_offset`, so an offset near the end resolves past it into the next block's source text.

Changed blocks therefore get `SourceInfo::Generated` (`by.kind = "nested-cell-mask"`), for which `map_offset` returns `None` — location unknown rather than a confident wrong answer, which `build_source_map` already tolerates. The origin anchor is `Other`, not `Invocation`: `preimage_in`'s `Generated` arm walks only `Invocation`, so the anchor is provably inert to any byte-copying writer.

**Accepted cost:** the writer's piece loop is top-level only, so a masked block nested in a div or list marks its top-level ancestor too — that container reads as location-unknown, unrelated prose inside it included. Unknown beats wrong. The marking lives only on the short-lived clone; the reparsed AST gets fresh `source_info`, so nothing `Generated` reaches the document.

## Scope

**In:** braced executable openers inside a markdown-displaying `CodeBlock` — attr classes empty or `["markdown"]`. Every braced opener, regardless of language.

**Out, deliberately:**

- **`RawBlock`** — a spike measured that masking a `{=markdown}` RawBlock converts "wrongly executed" into *silently mangled*: the writer emits those unfenced, so the inner cell becomes a genuine top-level cell. A downgrade, not a fix (bd-4gzwls31).
- **Doubled-brace `{{lang}}`** — no scanner executes them; that is a diagnostics gap (bd-q250-nested-fence-blind-spot-t68z1lsw).
- **`{=html}`-style openers** — only `breakQuartoMd` matches `=`, and it mis-*partitions* rather than executes.
- **A cell nested inside another executable cell** — rewriting there would corrupt code that is about to run.
- **` ````qmd `** — display classes stay empty-or-`markdown`; Q-2-50's hint names `markdown`, and widening is a separate decision.

Also corrected during implementation: a 4-space indented display block is **not** a second route to the empty-classes AST shape. qmd has no indented-code-block production at all — the scanner emits `INDENTED_CODE_BLOCK_DISALLOWED` and never produces a `CodeBlock`.

**Known limitation:** unmask pattern-matches, so an author who writes `q2-nested-executable` verbatim inside a display block gets it rewritten. Measured, negligible, and removed later in the epic when the assembler restores from retained bytes instead of matching. Every mechanism here is replaced rather than contradicted by later epic work.

## Docs

`docs/guides/authoring/computations.qmd` already asserted the behaviour this bug breaks — *"the inner cell is shown exactly as written (single braces and all) and is not executed"* — above a body that read `TBD.`. Now expanded with the rule, the nesting requirement, cell options in the fence header, deeper nesting, and the Quarto 1 migration callout.

**The page deliberately gets no live executable cell.** `docs/` is rendered in CI in exactly one place (`release.yml`, via `build-agents-docs`), no workflow installs R, Python, Jupyter or Julia, and an unavailable engine is a hard error rather than a fallback — so a live cell would fail the *release* build rather than a PR check, and `release.yml` already records that a hard docs-render failure cost a release. The regression tests carry the live case, gated on knitr.

## Tests

31 new tests across four tiers, each of the plan's 22 seam rows bound before dispatch to the production hunk whose revert turns it RED:

- 17 unit- and writer-tier in the module
- 9 render-tier through `render_document_to_file`
- 4 capture-tier at the engine-execution, preview-record, and hub-provider seams
- 1 in the jupyter text engine

Two rows are green-at-baseline **guards**, labelled as such rather than passed off as REDs: T3 (a real cell still executes) and the render-tier control (a display block in a document with no live cell renders unchanged). The bug only under-masks, never over-masks, so both already hold with zero masking in place — they guard against a regression the fix could introduce.

## Verification

- `cargo xtask verify` (full, no skip flags) — all 14 steps passed, run on this branch as rebased onto `b93c62c74`
- its workspace test run — **13621 passed, 199 skipped, 0 failed** (480.7s)

The delta is accounted: the plan's Phase 0 measured a live baseline of 13550; `main` has since added 40 tests (`f14cd4963` +12, `9ad1d4420` +28), and this branch adds 31, giving 13621 exactly. No stray or duplicated tests. The hub-build leg also left the worktree clean, reproducing the committed sandboxed-preview bundles byte-for-byte.

The full verify is what caught the one dependency bug: `quarto-core` declared `regex` only under `cfg(not(target_arch = "wasm32"))`, so the WASM preview pipeline — which must mask displayed cells too — would not compile. `--skip-hub-build` never compiles `quarto-core` to wasm32 and would have shipped that undetected.

End-to-end output was inspected per CLAUDE.md; the exact invocations and observed HTML excerpts are recorded in the plan, including the two discriminators that separate a masked display block from an executed one.

## Not in this PR

The bundle-rebuild commit that `cargo xtask verify`'s hub-build leg produced along the way is unrelated to this strand and went straight to `main` instead: `546fdd374` + `b93c62c74`. It corrects two checked-in artifacts that were already wrong there — `serviceWorker.js` still carried the offline cache that `103af4445` commented out of its source, and `q2-sandboxed-preview.html` embedded KaTeX 0.18.2 although the lockfile pins 0.18.4 (#637 bumped it; `6549a98ae` rebuilt against a stale `node_modules` and reverted it).

## Plan

`claude-notes/plans/2026-09-02-nested-cell-masking.md` is reconciled: every phase ticked, each `- [x]` re-checked against the tree rather than trusted, and the Phase 7 findings recorded. The outcome is commented on the strand.

### DIFF
--- a/claude-notes/plans/2026-09-02-nested-cell-masking.md
+++ b/claude-notes/plans/2026-09-02-nested-cell-masking.md
@@ -541,9 +541,39 @@ Phase 1 cannot pin this — a correct-looking render is achievable with the
 provenance wrong, so an implementer who skips this phase gets a green suite
 and ships the drift. These tests are the only thing guarding it.
 
-- [ ] Changed blocks, and the top-level ancestor of a changed nested block, get `Generated` + the `Other("nested-cell-mask/origin")` anchor
-- [ ] Test: `map_offset` into a masked block returns `None`; an unmasked sibling in the same document still resolves to its true offset
-- [ ] Test: a document with no nested fences yields a `SourceInfo::Concat` whose pieces are identical to an unmasked run (compare piece `offset_in_concat`/`length` pairs)
+- [x] Changed blocks, and the top-level ancestor of a changed nested block, get `Generated` + the `Other("nested-cell-mask/origin")` anchor
+- [x] Test: `map_offset` into a masked block returns `None`; an unmasked sibling in the same document still resolves to its true offset
+- [x] Test: a document with no nested fences yields a `SourceInfo::Concat` whose pieces are identical to an unmasked run (compare piece `offset_in_concat`/`length` pairs)
+
+**Phase 3 findings (2026-09-03).** All stamping lives inside `mask` (R1) — no
+second entry point. Two call sites produce the `Generated { by: { kind:
+"nested-cell-mask" }, from: [Other("nested-cell-mask/origin")] }` node: (a)
+`mask_block`'s direct-`CodeBlock` arm stamps the block itself the moment
+`mask_code_block_text` reports a change; (b) `mask`'s top-level loop captures
+each top-level block's pre-mutation `SourceInfo` before recursing, and — only
+if that top-level block is not already `Generated` from (a) — stamps it too
+after `mask_block` returns `true`. This covers both shapes the writer's
+top-level-only piece loop needs to see: a display block that *is* the
+top-level block (T15), and one nested inside a container whose top-level
+ancestor must carry the marking instead (T17), without double-stamping the
+same-object case. A private `mark_generated` helper (not a public API)
+factors the shared construction.
+
+Verified (`nested_cell_mask.rs` only touched; `cargo clippy -p quarto-core
+--all-targets -- -D warnings` clean): `quarto-core` 4329 run / 4319 passed /
+10 failed / 31 skipped — down from Phase 2's 12 failures by exactly T15 and
+T17, the Phase 3 gate. The remaining 10 (T1, T2, T4–T8, T19, T20, T22) are
+render/capture-tier, expected red pending Phase 4's seam wiring. T3 and the
+render-tier control (over-masking guards) reconfirmed still green, as are T16
+and T18 (Phase 2's writer-tier guards against over-application).
+
+Workspace boundary: `cargo nextest run --workspace` → 13581 run / 13570
+passed / 11 failed / 199 skipped. Delta against the Phase 2 boundary (13581
+run / 13568 passed / 13 failed / 199 skipped): +2 passed, −2 failed, run and
+skipped counts unchanged — exactly T15 and T17 flipping green, no other test
+moved. The 11 remaining failures are precisely the expected-red set carried
+forward: T1, T2, T4–T8, T19, T20, T22 in `quarto-core`, plus T21 in
+`quarto-hub-provider` (Phase 4's job).
 
 ### Phase 4 — Wire the seam
 
--- a/claude-notes/plans/2026-09-02-nested-cell-masking.md
+++ b/claude-notes/plans/2026-09-02-nested-cell-masking.md
@@ -577,13 +577,43 @@ forward: T1, T2, T4–T8, T19, T20, T22 in `quarto-core`, plus T21 in
 
 ### Phase 4 — Wire the seam
 
-- [ ] Mask the clone inside the per-engine loop, before `serialize_ast_to_qmd`; unmask `result.markdown` before the capture emit
-- [ ] `compute_input_qmd` masks; `write_review_file` unmasks before writing the consent artifact
-- [ ] Test: `compute_input_qmd` bytes equal `capture.input_qmd` for a document **containing a display block** — the gap in the existing invariant fixture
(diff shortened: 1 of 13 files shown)


gordonwoodhull:
Needs final revision and it will have a Q-2-50 refinement stacked on it.

Issue #342, opened 2026-06-25
local-first flow into quarto-hub.com


cscheid:
@vezwork, you have the vision here, please write it out :)

---

When they visit quarto-hub.com for the first time (and subsequent times) people should be able to use quarto-hub.com as a static (offline) app without signing in.  (indexxeddb only storage, no sync server connection)

People will be able to create projects, edit them, preview; generally take advantage of all the front end features of the app without logging in. Once they want to start syncing and sharing, they can choose to sign in. This is an onboarding ramp into signing up for the full syncing(+other server side stuff) features, still giving users good utility, and not costing us almost anything. This is a unique opportunity because we have a mostly local first architecture already, enterprise SaaS apps cannot do this. 

Requires two changes:
1. Remove auth gating, moving auth gating to behind a "I want to start syncing / start sharing" button or something
2. Local project set v.s. central project set. Carlos wants the sync server to store a centralized `user id -> project set` map, for convenience of having the same projects across browsers

cscheid:
They wouldn't get collaborative editing though, right?

vezwork:
Yea, no collaborative editng until people choose to sign up. We will want people to sign up and agree to terms and conditions and understand privacy policy and all that before they start collaborating (via our sync server) anyway. This makes the process a bit more granular and allows the users to try out stuff before committing to signing up.

PR #678, opened 2026-09-11
Say what Quarto Hub is on the quarto-hub.com landing page

icarusz:
`LoginScreen` is the quarto-hub.com landing page — App renders it whenever auth is enabled and absent — and it showed a logo, the words "Quarto Hub", "Sign in with Google to continue", and the GIS button. A first-time visitor, including someone arriving from an invite link, learned nothing about what they were signing in to.

The quarto-hub.com landing page

Both "Learn more" links — this page's and the invite cards' — now point at the github pages site instead of the provisional `quarto.org` placeholder (closes bd-rh2n4d7q) and open in a new tab: following one in place abandons the sign-in the visitor came to complete, and on an invite the invite link with it.

Two dark-mode defects fixed on the way

## Testing

Tests written and verified failing first, per `CLAUDE.md`.

`e2e/landing-card.harness.spec.ts` is new. It measures **rendered geometry** — glyph positions, line-box counts, box ratios — rather than the CSS properties that produce them, because the card's alignment is deliberately *mixed*: centered masthead and CTA, left-aligned prose. Centering the description was considered and rejected; it is the one block a first-time visitor has to read, and centering moves the start of every line. Measuring geometry means the mix survives a later "make this consistent" edit and fails on what a reader would actually notice.

Two cases in `invite-landing-theme.harness.spec.ts` pin card-over-page elevation. The load-bearing half is the **direction** — 1.06:1 clears any magnitude floor, so a contrast threshold alone would have passed the shipped bug.

- `npm run test:ci` — 1195 unit, 131 integration, 133 wasm
- full harness config — 185 passed, both themes
- `npm run build:all`, `npm run typecheck`, `npm run lint:css` — clean

Verified in a real browser against `vite dev`: all three landing states render, and every Learn more across both surfaces resolves 200 with no redirects.

Plan: `claude-notes/plans/2026-09-11-landing-page-intro.md` (records the 15 decisions, including the ones that reverse earlier ones).

Braid: bd-g0uyp2v1


vezwork:
## Questions on Project Sets

Does this now automatically create a new project set if you don't have one? If we merge this I don't think we will have a way to have the same project set on multiple devices/browsers any more. The previous screen allowed you to paste in a project set ID. 

Should we add a new way to change project sets? If you already have some collections/projects, perhaps changing project sets should merge your collections and projects from the project set you are in to the project set you are changing to?

## Note on Auth Gating

I'm still keen to [remove auth gating from the app](https://github.com/quarto-dev/q2/issues/342) and make it so people can start making projects and previewing docs (without syncing) immediately, and later sign in once they want to start sharing & collaborating. Therefore, I prefer there to be no landing page at all. That being said, in the meantime, this is definitely a less confusing landing page than the previous one.

hub-client/src/strings.ts
```diff
@@ -110,6 +110,44 @@ export const sections = {
...
+ * repeated: that a project link needs no account or Quarto install. That
+ * is the product's intent, but this deployment is allowlisted and the
+ * reader is looking at a sign-in button, so it would contradict the page
+ * it sits on.
+ */
+export const landing = {
+  product: 'Quarto Hub',
+  /** Rendered as two lines: the second sentence is the turn. */
+  taglineLead: 'Prose and code belong in one place.',
+  taglineFollow: 'So do the people.',
```
cscheid, line 134: I'm still not in love with "the people" here, it feels a bit too much like The People.

"So does all your team"? "So do your collaborators"?

hub-client/src/strings.ts
```diff
@@ -110,6 +110,44 @@ export const sections = {
...
+ * describe the product the same way. One claim there is deliberately not
+ * repeated: that a project link needs no account or Quarto install. That
+ * is the product's intent, but this deployment is allowlisted and the
+ * reader is looking at a sign-in button, so it would contradict the page
+ * it sits on.
+ */
+export const landing = {
+  product: 'Quarto Hub',
+  /** Rendered as two lines: the second sentence is the turn. */
+  taglineLead: 'Prose and code belong in one place.',
```
cscheid, line 133: "the same place"?

cscheid:
Ok, you know what? We need this more than we need the tweaks. We can improve it later.

PR #689, opened 2026-09-17
Self-heal a permanently-stuck index document after an actor-id collision (bd-6f21d4c6)

gordonwoodhull:
## What's actually wrong

`actorIdFromUserId` (`hub-client/src/services/userSettings.ts`) deliberately derives a *stable* Automerge actor id from the local `userId`, reused across every session/tab for that user — "instead of a fresh random actor each session," per its own doc comment. That's fine in isolation, but it means two live sessions for the same person (two tabs, a stale tab plus a fresh reconnect) share one actor id. If both commit a local edit to the same document before either has synced the other's change, both edits can end up claiming the same `(actor, seq)` pair — a genuine identity violation, not a CRDT merge conflict.

When that happens, `automerge` correctly throws `RangeError: duplicate seq N found for actor <id>` out of `receiveSyncMessage`. `automerge-repo`'s own `Repo#receiveMessage`, however, catches it with:

```ts
this.synchronizer.receiveMessage(message).catch(err => {
  console.log("error receiving message", { err, message })
})
```

— a bare `console.log`, no retry, no escalation. Because the throw happens before either side's sync state advances, the hub keeps resending the identical colliding change on every subsequent attempt, and the affected client keeps rejecting it, identically, forever. This is not "a client that's a bit behind" — it's a permanent wedge, and the only known workaround was manually clearing the browser's `automerge` IndexedDB database, which nobody would think to automate without knowing the mechanism.

Confirmed directly against the vendored `automerge`/`automerge-repo` v2.5.6 source and reproduced end-to-end: two real `SyncClient`s sharing one actor id, racing a local edit each against a real in-process hub, hit this exact swallowed catch and go permanently stuck.

## What this PR does — and doesn't do

It gets an already-stuck client **unstuck**. It does not stop the collision from happening in the first place — that's a separate, not-yet-attempted fix (either give `actorIdFromUserId` a per-session-unique component, or contribute upstream, where the right primitive — `RepoConfig.actorIdPrefix` — was prototyped in 2024 and never shipped). Scope is deliberately narrow: **index documents only**, not file documents, since the reported symptom is specifically an `IndexDocument` going stale (Q-13-4 squiggles on links to files that genuinely exist).

`installDuplicateSeqRecovery` wraps `repo.synchronizer.receiveMessage` — a public method on `Repo`'s own public `synchronizer` field — at the *instance* level for the connection's own `Repo`. It calls through to the original method and taps its rejection with an additional `.catch()`, so the existing swallow still fires unchanged; nothing in `automerge-repo` is edited, and no other `Repo` instance is affected. On a match (`message.documentId` is this connection's index doc, error is a `RangeError` matching `/duplicate seq \d+ found for actor/`), `recoverIndexDocument` runs the automated, scoped equivalent of the manual IndexedDB wipe, entirely through public API:

- `repo.delete(indexDocId)` — confirmed against source to drop the `DocSynchronizer`'s wedged per-peer `SyncState` for that one document *and* purge its persisted storage entry, leaving every other document untouched.
- `findDoc<IndexDocument>(indexDocId)` (the same retry-aware helper `connect()` already uses) forces a clean full resync from the hub instead of replaying the poisoned local history.
- Swaps in the fresh handle, re-subscribes, and reconciles file/identity/capture state against whatever changed while the client was stuck.

A guard mirrors an existing race-defense already in `findDoc`'s retry loop: if the connection was torn down or replaced while the re-fetch was in flight, the stale result is discarded rather than applied to unrelated new state.

## Why this is safe to ship without touching `automerge-repo`

The monkey-patch is an own-property override on one `Repo` instance (`synchronizer.receiveMessage = wrapped`), not a class/prototype edit — `CollectionSynchronizer.prototype.receiveMessage` in `node_modules` is untouched. If `repo.synchronizer` isn't the expected shape (a few existing tests mock `Repo` without one), installation is a silent no-op rather than a crash — this is a best-effort enhancement, not a required invariant for `connect()` to work.

## Test plan

- `full-stack-actor-collision.test.ts`: two real `SyncClient`s share one actor id against a real `startTestHub()`. Each commits its own `createFile()` synchronously (no `await` between the two calls, so neither can process the other's message first — plain JS run-to-completion semantics, no fake-timer trickery needed). Asserts, in order: the collision + swallowed catch really happen; the fix's recovery log fires; a subsequent edit from an uninvolved third client (different actor id) reaches **both** previously-colliding clients within a bounded window; a second later edit also reaches both, ruling out "one lucky retry."
- Verified **red before green**, per this repo's TDD rule: temporarily commented out both `installDuplicateSeqRecovery` call sites, confirmed the test failed exactly as expected (collision still fires, no recovery log, full poll timeouts consumed), then restored and confirmed green. Stable across repeated runs (~0.4–0.6s each).
- `actor-id-collision.test.ts` independently confirms the underlying collision at the bare-automerge level (no `automerge-repo`, no network, no hub) — the fact the fix's detection regex depends on.
- Full package suite: 140/140. `npx tsc --noEmit` clean in both `quarto-sync-client` and `hub-client` (which bundles this package from source).

Design writeup with both decision points (detection mechanism: wrapping `receiveMessage` vs. sniffing `console.log`; recovery scope: index-doc-only vs. generalized) and the alternatives considered: `claude-notes/plans/2026-09-17-index-doc-duplicate-seq-self-heal.md`.


gordonwoodhull:
When I asked if it was because we are using actor ids wrong:

> Automerge's actor-id contract assumes one actor id authors a single linear sequence of changes — it's meant to be an ephemeral, per-writing-session identifier, not a durable identity. Automerge's own error for this exact situation is literally named `DuplicateActorId` with the message "possible document clone" — the library's own vocabulary frames "same actor id, two concurrently-edited copies" as a recognizable misuse pattern, not just an edge case.      

> `actorIdFromUserId`'s own doc comment is explicit that it's a deliberate departure from that norm: "instead of getting a fresh random Automerge actor each session," it derives a stable id from userId specifically to get durable per-user attribution (so identity/presence features can key off a  consistent id across a user's sessions). Automerge's default model doesn't give you that for free — random-per-session actor ids are safe but don't carry identity.

> So it's not a case of the code doing something Automerge accidentally allows and nobody noticed — it's a real tension between two things Automerge doesn't offer simultaneously (safe concurrency vs. stable identity), resolved in favor of stable identity, with the concurrency risk apparently not fully worked through at the time. Notably, upstream automerge-repo maintainers recognized the same need and started building a primitive for it (a `stable-actor-ids` branch adding actorIdPrefix, combining a stable prefix with a per-session-unique suffix — exactly the shape that would fix this) but never merged or shipped it. That suggests the need was legitimate and the current approach wasn't obviously wrong at the time it was written — it's more that q2 reached for stable identity
before the ecosystem had a blessed, safe way to get it.

Monkey-patching automerge to add error handling is kind of ugly, but there should be a proper fix ahead, and we can remove this when it's no longer needed.

PR #702, opened 2026-09-21
Replace CLAUDE.md with AGENTS.md

shikokuchuo:
## Summary

- Move the project instructions from `CLAUDE.md` to `AGENTS.md`. Claude Code reads `AGENTS.md`, the same as other agents.
- Rename the per-crate instruction files (`pampa`, `wasm-qmd-parser`, `wasm-quarto-hub-client`) to `AGENTS.md`.
- Update all references in comments, skills, and notes.

No code changes. Comment and documentation text only.

## Test plan

- `cargo build --workspace` and `cargo xtask lint` pass.
- A repo-wide search confirms that no stale `CLAUDE.md` references remain. Two groups of references stay on purpose: the filename detection in `project/discovery.rs` (user projects can still use `CLAUDE.md`), and historical notes in `claude-notes/`.


cderv:
Which is the minimum claude code version for this ? So we're sure it works for us too :)

shikokuchuo:
[v2.1.277](https://code.claude.com/docs/en/changelog#2-1-277) - we're already on v2.1.278.

cscheid:
This is good but it changes all of our ongoing agent sessions, etc, so let's warn everyone about it on Wednesday and come up with a time to do it.

cscheid:
Let's hold off on this for a little bit: https://blog.szypowi.cz/p/claude-code-reads-agents.md-only-when-telemetry-is-on/

cscheid:
Looks like this will be fixed [soon](https://news.ycombinator.com/item?id=49815363).

cderv:
While learning a bit more about this, one caveat that seems to exists

if a CLAUDE.md ever appears in a parent directory of this repo, Claude Code will read that instead and ignore the AGENTS.md here. Worth keeping in mind depending on what you have in your local organization. 

Maybe there is a claude config to avoid this. 🤷

PR #704, opened 2026-09-21
Support docx, pptx, epub, and typst output via Pandoc writers

gordonwoodhull:
## Summary

Quarto can now render `docx`, `pptx`, `epub`, and `typst` documents by handing its own document representation to Pandoc's native writers, rather than routing them through Quarto's own HTML-only rendering path. To do this, Quarto feeds its custom AST nodes — crossrefs, callouts, theorems, tabsets, floats — over a documented wire format to a vendored copy of Quarto 1's Lua filter pipeline, which turns them into the constructs each Pandoc writer understands.

Trigger a format the same way as any other: `format: docx`, `format: pptx`, `format: epub`, or `format: typst` in front matter or `_quarto.yml`, or `--to <format>` on the command line. `docx`, `pptx`, and `epub` shell out to a `pandoc` binary on `PATH` (or one pointed to by `QUARTO_PANDOC`); `typst` shells out the same way to a `typst` binary (or `QUARTO_TYPST`) and compiles the intermediate `.typ` source to a PDF.

## What carries over from Quarto 1

- **Cross-references and numbering** — `@fig-`, `@tbl-`, `@eq-`, and theorem-family refs resolve and number correctly in every new format. Typst numbers natively through its own compiler rather than through Quarto's numbering pass, so the two always agree.
- **Callouts and theorems** (`callout-note`, `.theorem`, proofs, etc.) render with their format-appropriate styling.
- **Tabsets** render their content in every format; EPUB and (where applicable) other non-interactive readers show all panels expanded rather than collapsing to one active tab, since there's no JavaScript to hide the rest.
- **Captions, figures, and images** — including image resource staging into the output container (e.g. `word/media/` in docx).
- **`reference-doc` / `template`** for docx/pptx, and custom `.typ` templates for typst, resolve the same way other Quarto paths do and fail with a clear diagnostic if the file is missing.
- **`brand.yml`** colors and typography apply to typst output the same way they apply to other formats.
- **Typst margin notes** via the vendored `marginalia` package, and automatic fetching of any additional Typst Universe package a document's raw Typst content references.
- **EPUB-specific options**: `epub-chapter-level`, `epub-cover-image`, `epub-metadata`, `epub-embed-font`, and an additional `css` stylesheet merged with Quarto's own callout/layout styles.
- **`content-hidden` / `when-format` conditional content** is honored for all of these formats, including content nested inside custom nodes.

## Rendering more than one format

If a document's front matter declares more than one `format:`, Quarto still renders only one target per invocation — but it now warns, naming which format it used and which it skipped, instead of silently dropping the others.

## Known gaps

- **`latex`/`pdf` is not yet supported** — `--to pdf` and `--to latex` still refuse to render.
- **Mermaid diagrams don't rasterize** in docx/pptx/epub: no image is embedded, though the code cell's source text is preserved in the output. (Quarto 1 renders these through a headless browser; the new pipeline doesn't yet invoke one.)
- **Pandoc-specific failures get their own diagnostics instead of a generic error.** New `Q-20-*` codes cover a missing `pandoc` binary, a version too old, a nonzero pandoc exit (with its stderr/filter traceback attached), and a couple of internal wire-format consistency checks.


cderv:
So cool @gordonwoodhull - Thank you!

